PR #14, opened 2018-03-05
All texts resized in vertical mode

Genetik57:
Credit : @Supernature2k
Link : https://gitlab.com/recalbox/recalbox-emulationstation/commit/875e235ed4cd3fdea722deebd9d6605f32e75661


nadenislamarre:
@Genetik57 which bug is it supposed to fix ? i don't reproduce a bug with vertical mode.

PR #15, opened 2018-03-05
Fix bug in OSK

Genetik57:
*Potential bug in array for OSK

Credit : @Supernature2k
Link : https://gitlab.com/recalbox/recalbox-emulationstation/commit/9f3f052ca769647588a6df232b8fb7180f8feec1


nadenislamarre:
@Genetik57 which bug is it suppose to fix ?
i never get a feedback about the keyboard not working.
can you reproduce a bug ?

PR #28, opened 2019-01-14
fix: Bezel

jdorigao:


nadenislamarre:
decorations->add(_("NONE"), "none", RecalboxConf::getInstance()->get("global.bezel") == "none");
must be out of the loop

PR #49, opened 2019-04-27
Add French translation

rptx:
Translate "FILES" into french.


nadenislamarre:
thanks a lot.
i will not directly merge this pr because i'm doing several changes and the pr will not apply on my current code. but while it is simple, i will translate it. thanks.

nadenislamarre:
done without the merge.

PR #60, opened 2019-05-14
Added a menu in "UI Settings" to install Batocera themes.

lbrpdx:
Requires /userdata/scripts/batocera-es-theme and Batocera 5.22+


nadenislamarre:
thanks a lot.
i thought to create it in system>update because the ui menu is already very full
but let's keep it like this for the moment.

PR #61, opened 2019-05-15
filterTrigger: generalize for more controllers

Genetik57:
* Filter axes 2/5 for all devices, but only if LeftTrigger or RightTrigger is being configured. This should fix compatibility with XBox 360 and other generic controllers that use these axes for triggers, but won't affect other controllers using these axes for analog sticks, etc.
* Improve third-party PS3 detection (some Shanwan controllers have a grave accent in place of 'm' for 'Gamepad').

Credit: https://github.com/psyke83
Source: https://github.com/RetroPie/EmulationStation/commit/93fdfaa9c2815edd34ed79883d5a983eefde17d1


nadenislamarre:
the filter function is a hack in retropie-es (disabled in batocera, see the return false on top of the function).
on batocera, we fix this kind of things lower, in sdl or with evdev rules, not at top application layer.

PR #62, opened 2019-05-15
Cosmetic modification

Genetik57:
*Several cosmetic modifications


nadenislamarre:
dephasing code with retropie-es code. merge would be more complicated.
eventuallly, push it on retropie-es.

PR #82, opened 2019-06-04
Fix: long theme names/URLs can crash the install MsgBox

lbrpdx:
Discovered this when installing https://github.com/sorkenbarriukan/ES-Theme-batocera-clasic-carousel/tree/master (nice theme btw!)


lbrpdx:
The crash that is fixed here can be reproduced when you add the following theme (from /userdata/system/themes.txt) with the previous code:
`Classic-carousel https://github.com/sorkenbarriukan/ES-Theme-batocera-clasic-carousel`

PR #90, opened 2019-06-20
Fix Kiosk Favorites not registering

Genetik57:
Had been broken a few months back

Source: https://github.com/RetroPie/EmulationStation/commit/1b06df1572fcdae21663aa138e1366d7d7714c42
Credits: @pjft


nadenislamarre:
i thought it was the wanted behavior in fact both behavior are ok

PR #97, opened 2019-06-30
Add Game Info Font Settings to OMX Player

Genetik57:
Adds the ability for users to change the following features in OMX Player's subtitles (game info for screen saver).

- font size
- font file path
- italic font file path
- subtitle position

These changes can be made in the GUI menus via the Video Screen Saver menu, or directly in the es_settings.cfg file.

Safe OMX Player defaults are hard-coded into the EmulationStation application.

Here's an example of the new keys loaded/saved into in the es_settings.cfg file:

- `<int name="SubtitleSize" value="38" />`
- `<string name="SubtitleFont" value="/usr/local/share/fonts/slkscr.ttf" />`
- `<string name="SubtitleItalicFont" value="/usr/local/share/fonts/slkscrb.ttf" />`
- `<string name="SubtitleAlignment" value="center" />`

*Source: https://github.com/RetroPie/EmulationStation/commit/a2f59ee2fcae35dbaad7264226c6a8541d7f9f4d
*Credits: @danovision


nadenislamarre:
i've doubts about this patch. can you confirm you compiled es with this patch and tested it ?

PR #107, opened 2019-08-09
MP3 titles pop-up window + fade out

lbrpdx:
Two features in this PR:
1. Rather than an abrupt stop, background MP3s are now fading out for 0.5 second before launching a game.
2. New audio config option (`audio.display_titles=1` in batocera.conf) that pop-up a window whenever a new background MP3 is played back. What is displayed in the pop-up is:
  2.1 If there's a ID3v1 tag, show the ID3v1 tag
  2.2 If no ID3v1 tag, look for an ID3v2 tag and show it
  2.3 If neither ID3v1 nor v2 are present, show the filename of the MP3.


nadenislamarre:
very nice.
i've not finnished the review, but i've 2 points for the moment :
- some // batocera are missing
- SystemConf::getInstance()->set("audio.display_titles", display_titles->getState() ? "1" : "0");
i think this is more a es config option more than a system option.

lbrpdx:
After reviewing the other options in the "Sound settings" menu, I can see that all the existing options are pointing to batocera.conf:
`audio.device
audio.volume
audio.bgmusic
audio.persystem`
For consistency, I think we should keep this new option in batocera.conf.

nadenislamarre:
ok. effectively, consistency is better.
"one day" we will change that

PR #113, opened 2019-08-15
Fix Transform4x4f which was broken in many, many ways, but just out o…

Genetik57:
…f pure luck worked anyway

Source: https://github.com/RetroPie/EmulationStation/commit/28168667ea9a0addd16dfb82253c4f5dfef0e927
Credit: @tomaz82

### DIFF
--- a/es-core/src/math/Transform4x4f.cpp
+++ b/es-core/src/math/Transform4x4f.cpp
@@ -8,21 +8,21 @@ const Transform4x4f Transform4x4f::operator*(const Transform4x4f& _other) const
 	return
 	{
 		{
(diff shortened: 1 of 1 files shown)


tcamargo:
@nadenislamarre This is a "cherry-pick" from retropie-es. How do you plan to handle that? I would prefer to merge our source tree with retropie's.

nadenislamarre:
yes. i want to merge after 5.23 is out instead of commit by commit.
i didn't merge yet because in the merge, you'll notice a hard coded path to a ttf, and i've not check how it handles on batocera (not checked if the file exists, i did the merge during my holliday out of my computer, just to check)

PR #120, opened 2019-08-22
Fixed AudioManager, classics Sounds and SongNameThread

fabricecaruso:
There were 3 potential crash sources :

Rewritted some parts of AudioManager to avoid init of SDL_Mixer ( with Mix_OpenAudio ) only when the music starts and close it when music stops, and make better cleanings. Init of SDL_Mixer is in Audiomanagers's init, so we can use SDL_Mixer for classic sounds too.

Rewritted Sound class to use SDL_Mixer instead of using SDL_Audio that was disabled.
I don't know if it could lead to **crashes** on linux, but it is still possible as the required audio engine is not running.

Deleted SongNameThread, I moved it to Window::update ( in the main UI loop ) because :

- When you launch a game, the thread is still running, the AudioManager has been deinit and set to null. The thread reinits the AudioManager after 3 seconds. When you come back from the game, Audiomanager is Inited again.
- Accessing songName variable in a thread and also in the UI thread make strings written at the same time and currupts adresses in memory - this can lead to a **crash** with access violation.
- Window::update is never called when a game is launched because system() is waiting for the running process to end.

I also removed this test to move the test in Sound.cpp
if (sInstance == nullptr) // !Settings::getInstance()->getBool("EnableSounds"))
If EnableSounds is false, AudioManager is always null. Talking to Audiomanager to start music makes an **INSTANT CRASH** ( null pointer exception )

Fixed sometimes indentation at 2 spaces with tabs, which is not the original convention in ES ( makes a lot of lines dirty )


es-app/CMakeLists.txt
```diff
@@ -93,8 +92,7 @@ set(ES_SOURCES
     ${CMAKE_CURRENT_SOURCE_DIR}/src/FileFilterIndex.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/src/SystemScreenSaver.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/src/CollectionSystemManager.cpp
-    ${CMAKE_CURRENT_SOURCE_DIR}/src/NetworkThread.cpp # batocera
```
tcamargo: Why do you removed NetworkThread?

es-core/src/Window.cpp
```diff
@@ -207,12 +208,21 @@ void Window::input(InputConfig* config, Input input)
 
 void Window::update(int deltaTime)
 {
-        // batocera
-        if(!mMessages.empty()){
+	// batocera        
+	if (SystemConf::getInstance()->get("audio.display_titles") == "1")
```
tcamargo: What about using mMessages?

es-core/src/Sound.h
```diff
@@ -2,7 +2,7 @@
 #ifndef ES_CORE_SOUND_H
 #define ES_CORE_SOUND_H
 
-#include "SDL_audio.h"
+#include "SDL_mixer.h"
```
tcamargo: Too many changes in standard ES core code... Why?

tcamargo:
@fabricecaruso Could you please detail a bit what are you correcting? I put some comments directly in the code.

es-app/CMakeLists.txt
fabricecaruso: No it's still there

es-core/src/Window.cpp
fabricecaruso: Still there

es-app/CMakeLists.txt
tcamargo: Look at the diff. You removed it from CMakeLists.

es-core/src/Window.cpp
tcamargo: Sorry?

es-core/src/Sound.h
fabricecaruso: Sound is a rewrite because It uses the SDL_Audio which is useless ( and don't work anymore ) -> I rewrited it to use SDL_Mixer instead as Music is using it too

fabricecaruso:
> @fabricecaruso Could you please detail a bit what are you correcting? I put some comments directly in the code.

Rewritted some parts of AudioManager to avoid ini of SDL_Mixer ( with Mix_OpenAudio ) only when the music starts and close it when it stops, and make better cleanings.
Rewritted Sound class to use SDL_Mixer instead of using SDL_Audio that was disabled.

Deleted SongNameThread, I moved to processing in Window::update ( in the main UI loop ) because :
- When you launch a game, the thread is still running, the AudioManager has been deinit and set to null. The thread reinits the AudioManager after 3 seconds. When you come back from the game, Audiomanager is Inited again.
- Accessing songName variable in a thread and also in the UI thread make strings written at the same time and currupts adresses in memory - this can lead to an access violation.
- Window::update is never called when a game is launched because system() is waiting for the running process to end.

es-core/src/Window.cpp
fabricecaruso: Still there, just moved after testing songname

es-app/CMakeLists.txt
fabricecaruso: NO !!! It's still there !!!!!!!!!!! Don't know why git consider it a change.

tcamargo: So, I can't merge you change. It would remove the line..

es-core/src/Window.cpp
tcamargo: yes... why dont use use it? fill mMessages? so, you just need to fill it with the songname and no changes are necessary here.

fabricecaruso: I didn't change previous lines of code found in SongNameThread, just moved them there

tcamargo: And I think it is not the right place. We should have used mMessages, if possible.

nadenislamarre:
i merge.
i build a version with it.
if it causes issues, we will revert.

PR #126, opened 2019-08-28
Set VSync on by default on Raspberry Pi

fabricecaruso:
On a TV with a Raspberry Pi, ES looks like it is flickering.
Set to true by default on every system.

### DIFF
--- a/es-core/src/Settings.cpp
+++ b/es-core/src/Settings.cpp
@@ -80,14 +80,7 @@ void Settings::setDefaults()
         mStringMap["INPUT P5"] = "DEFAULT";
         mStringMap["Overclock"] = "none";
 
-        // batocera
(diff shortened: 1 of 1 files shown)


fabricecaruso:
> ...but the comment in the code says RPis have trouble rendering with vSync on?
> I wonder, how do you assume most of the users plug their system on a TV?

It's a very old comment from Recalbox. Look at retropie's ES code (same file same place) they use vsync too now on the Pi.

fabricecaruso:
The best explanation is seeing ES running on a Pi3...

fabricecaruso:
> Well, the change will affect all RPi, from 0 to 3...

I have no way to test this setting in a pi0, but it should not affect performance. 
Es is not the problem on a pi0 ;-)

PR #128, opened 2019-08-29
AudioManager support & fixes with Navigation sounds

fabricecaruso:
- Fix a crash when returning from a game when "EnableSounds" is true
- Restore option "ENABLE NAVIGATION SOUNDS" in menus.


nadenislamarre:
conflict, can you rebase please ?

PR #132, opened 2019-08-30
Themable menus & Full support for RB Themes

fabricecaruso:
- Full support for menu themings & help images
- Full support for recalbox themes & every existing EmulationStation Themes.
- Add video & marquee support in detailed view (video is used in RB detailed view)
- Fixes for video start delay & fadings.
- Updated menu & buttons resources : they need to be white to draw correctly with themed menus.
- GuiMenu opening animation.
- Menu navigation : when on top & moving top, select last item ( et inversement... )


nadenislamarre:
you keep : current.name = "atari5200"; ?

fabricecaruso:
> 
> 
> you keep : current.name = "atari5200"; ?

It is a "hack" for windows in debug mode rounded with #ifdefs : I couldn't load bios information, so I wrote this to see what the menus looked like. This code can be removed in the future it is there for testing purposes only.

PR #158, opened 2019-09-13
Grid : add grid size override & change view per gamelist & misc fixes

fabricecaruso:
- Grid : add grid size override + custom view & grid size per gamelist
- Misc fixes ( keyboard, theme downloader, screenscraper when no game found... )
- Added option SHOW CONTROLLER ACTIVITY
- GPI Fixes

### DIFF
--- a/es-app/src/ApiSystem.cpp
+++ b/es-app/src/ApiSystem.cpp
@@ -1042,7 +1042,7 @@ std::vector<std::string> ApiSystem::getBatoceraThemesList()
 
 #if WIN32 && _DEBUG
 	// http://batocera-linux.xorhub.com/upgrades/themes.txt
-	res.push_back("[A]\tAlekfull\thttps://github.com/jdorigao/es-theme-alekfull");
+	res.push_back("[A] Alekfull - https://github.com/jdorigao/es-theme-alekfull");
 	res.push_back("[A]\tArt-book\thttps://github.com/anthonycaccese/es-theme-art-book");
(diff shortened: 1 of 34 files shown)


shantigilbert:
After this merge I now have these errors: 
```
../es-core/src/utils/StringUtil.cpp: In function 'std::vector<std::basic_string<char> > Utils::String::splitAny(const string&, const string&)':
../es-core/src/utils/StringUtil.cpp:333:9: error: 'strcpy' is not a member of 'std'
    std::strcpy(str, s.c_str());
         ^~~~~~
../es-core/src/utils/StringUtil.cpp:333:9: note: suggested alternative: 'string'
    std::strcpy(str, s.c_str());
         ^~~~~~
         string
../es-core/src/utils/StringUtil.cpp:335:16: error: 'strtok' was not declared in this scope
    char* pch = strtok(str, seperator.c_str());
                ^~~~~~
../es-core/src/utils/StringUtil.cpp:335:16: note: suggested alternative: 'strtoq'
    char* pch = strtok(str, seperator.c_str());
                ^~~~~~
                strtoq
```

Which should in theory be fixed by adding `#include <cstring>` but after I add that I get this error instead;

```
../es-core/src/LocaleES.h:8:38: error: no matching function for call to 'gettext(const std::basic_string<char>&)'
 #define _(A) boost::locale::gettext(A)
```

fabricecaruso:
Sorry for this, 
I've just pushed another (tested on ubuntu) PR to fix my mistakes.

PR #187, opened 2019-09-24
Game specific configuration : change naming convention

fabricecaruso:
Used to avoid conflicts between same rom name in different system + avoid problems reading batocera.conf when = or # caracters are found.

See FileData::getConfigurationName() for new naming rules which are :
systemname.filename (without spaces without = without #)

This PR is to merge only when script are to be adapted.

### DIFF
--- a/es-app/src/FileData.cpp
+++ b/es-app/src/FileData.cpp
@@ -40,6 +40,15 @@ const std::string FileData::getPath() const
 	return Utils::FileSystem::resolveRelativePath(mPath, getSystemEnvData()->mStartPath, true);	
 }
 
+const std::string FileData::getConfigurationName()
+{
+	std::string gameConf = Utils::FileSystem::getFileName(getPath());
+	gameConf = Utils::String::replace(gameConf, "=", "");
+	gameConf = Utils::String::replace(gameConf, "#", "");
+	gameConf = getSourceFileData()->getSystem()->getName() + std::string("[\"") + gameConf + std::string("\"]");
+	return gameConf;
+}
+
 inline SystemEnvironmentData* FileData::getSystemEnvData() const
 { 
 	return mSystem->getSystemEnvData(); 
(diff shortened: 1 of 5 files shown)


Darknior:
Thanks for this update, i 've report this future possible problem :p 
Is it not a performance problem if all the "games config" are stored in the same unique file ?

nadenislamarre:
it will create things like :

snes.emulator = libretro
snes.mariokart.smc.emulator = libretro
snes.mariokart.smc-renderer.shaders= scanlines

is it possible in es to change by :
snes["mario kart.smc"].emulator = libretro ?
snes-renderer["mario kart.smc"].shaders= scanlines?

maybe it is more readable ?
personnally, i prefer. don't you ?

in configgen (scripts), there is only 2 lines to change : https://github.com/batocera-linux/batocera.linux/blob/master/package/batocera/core/batocera-configgen/configgen/configgen/Emulator.py#L26

gameSettings = recalSettings.loadAll(os.path.basename(rom))
and
gameSettings = recalSettings.loadAll(os.path.basename(rom) + "-renderer")

by
gameSettings = recalSettings.loadAll(self.name + "." + os.path.basename(rom).translate({ord(i): None for i in '#= '}))
gameSettings = recalSettings.loadAll(self.name + "." + os.path.basename(rom).translate({ord(i): None for i in '#= '}) + "-renderer")

or         gameSettings = recalSettings.loadAll(self.name + "[" + os.path.basename(rom).replace("\"", "") + "]"
         gameSettings = recalSettings.loadAll(self.name + "-renderer" + "[" + os.path.basename(rom).replace("\"", "") + "]"

if we prefer the later solution (in which only " will be removed)

fabricecaruso:
>  snes-renderer["mario kart.smc"].shaders= scanlines?

Easy for standard settings, but not for this one, which should requires code refactoring to work :

Without refactoring, the result would be :
snes["mario kart.smc"]-renderer.shaders= scanlines?

( See GuiMenu.cpp -> SystemConf::getInstance()->set(configName + "-renderer.colorization", colorizations_choices->getSelected()) )

Agree with that ?

PR #195, opened 2019-09-28
Windows automated build : disable boost for msvc

fabricecaruso:
With that, an automated build for Windows is now available at : 
https://github.com/fabricecaruso/batocera-emulationstation/releases


fabricecaruso:
> Why such change for the buildroot Linux EmulationStation code? Anyway, does the code compile on Windows without boost?
Why ? To provide a windows build for themers or people wanting to "save" their batocera on windows...
Windows build does not use Boost. Linux version uses boost only for localisation.

PR #215, opened 2019-10-12
Updated mame games list

AlexxandreFS:
Updated the game display list for the latest stable 0.214 version of mame.
Maintained backward compatibility with name2003plus, mame2010 and FBANeo.


tcamargo:
Should we add those files to the lr-mame package?

AlexxandreFS:
Not! It is designed to display a friendly name on the Mame and NeoGeo list of roms on Emulationstation.
Example: The rom folder has a rom named mk.zip and the emulationstation list shows Mortal Kombat.

tcamargo:
Where those files come from?

AlexxandreFS:
I created it.
I exported the dat files from all versions of mame supported by batocera and FBA, then I learned how the original file structure worked and rebuilt the files with the most up to date information.

PR #217, opened 2019-10-14
Removed "auto" from shaders, new default is "none".

lbrpdx:
The shader set "auto" is removed as it's basically the same as "none" now -- see https://github.com/batocera-linux/batocera.linux/pull/987 
So, this PR removes the option and ensures a graceful transition from "auto" shader sets to "none" from Batocera 5.23 and earlier versions.


nadenislamarre:
no, this is not the same.
auto means that the user wants the batocera defaults. none, that he won't shaders.
for the moment it is none, but it can change, and this day we may to add none back.
moreover, sometimes, for other configuration, auto mean none or not depending on the board for performances reasons.

PR #237, opened 2019-10-26
Theming Fixes

fabricecaruso:
Fix : Theme subset selection can be empty & lead to a crash
Fix : VERTICAL_WHEEL carousel logo position on some themes using rotation by logo size + ensure loading order for themes <= 4
Fix : Rating Component drawing on some themes.
Fix : ImageComponent clipping with rotation


nadenislamarre:
/recalbox/share_init/system/.emulationstation/themes is not a valid directory
the 2 only correct values are :
/userdata/themes
and
/usr/share/emulationstation/themes

PR #240, opened 2019-10-30
Fixes : Folder navigation & sortings

fabricecaruso:
- Added ".." item when navigating in a subfolder.
- Faster case insentisive sorting + refactored sort calls.
- Show labels on grid items using default folder/game image, when labels are not defined by the theme.
- Added option "SHOW FILENAMES IN LISTS" to display filenames instead of scrapped names ( except arcade systems using mamenames )

### DIFF
--- a/es-core/src/components/ImageGridComponent.h
+++ b/es-core/src/components/ImageGridComponent.h
@@ -819,10 +819,10 @@ void ImageGridComponent<T>::updateTileAtPos(int tilePos, int imgPos, bool allowA
 
 		if (ResourceManager::getInstance()->fileExists(imagePath))
 			tile->setImage(imagePath);
-		else if (mEntries.at(imgPos).object->getType() == 2)		
-			tile->setImage(mDefaultFolderTexture);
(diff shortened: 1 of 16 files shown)


es-core/src/components/ImageGridComponent.h
```diff
@@ -819,10 +819,10 @@ void ImageGridComponent<T>::updateTileAtPos(int tilePos, int imgPos, bool allowA
 
 		if (ResourceManager::getInstance()->fileExists(imagePath))
 			tile->setImage(imagePath);
-		else if (mEntries.at(imgPos).object->getType() == 2)		
-			tile->setImage(mDefaultFolderTexture);
+		else if (mEntries.at(imgPos).object->getType() == 2 || (mEntries.at(imgPos).object->getType() == PLACEHOLDER && mEntries.at(imgPos).object->getPath() == ".."))
```
shantigilbert: Does not compile on Linux

```
../es-core/src/components/ImageGridComponent.h:837:100: error: 'PLACEHOLDER' was not declared in this scope
   else if (mEntries.at(imgPos).object->getType() == 2 || (mEntries.at(imgPos).object->getType() == PLACEHOLDER && mEntries.at(imgPos).object->getPath() == ".."))
```                                                                                                    ^~~~~~~~~~~

PR #241, opened 2019-10-31
LibRetro AI service for game translation.

lbrpdx:
This is the ES configuration screen for https://github.com/batocera-linux/batocera.linux/pull/1024   

Screen Shot 2019-10-30 at 7 25 26 PM

It basically manages the three new fields that have been added in `batocera.conf`:
```
  global.ai_service_enabled=1  # Is the service enabled?
  global.ai_service_url=       # URL of the translation service to use
  global.ai_target_lang=Fr     # Language to translate into (2-char language code)
```

### DIFF
--- a/es-app/src/guis/GuiMenu.cpp
+++ b/es-app/src/guis/GuiMenu.cpp
@@ -983,39 +983,37 @@ void GuiMenu::openGamesSettings_batocera()
 				SystemConf::getInstance()->get("global.ai_service_enabled") == "1");
 			ai_service->addWithLabel(_("ENABLE AI SERVICE"), ai_service_enabled);
 
-			// Target language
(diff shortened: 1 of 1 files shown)


Genetik57:
Amazing I can not wait to see him

lbrpdx:
Hum... i thought about this and here is my thought: as the API key is part of the URL, it would be more friendly to have this defined in ES than configgen. But I don’t want to have a new field just for the API key that simply is « BATOCERA » as I can anticipate some users will change it with their own string... and break it, so I believe it’s better to make it part of the URL. I don’t think we want to have API keys hardcoded in configgen really. 

I am not sure I understand your second point. Are you talking about a #define for the URL?

Darknior:
Fine for me like we talk about it on Discord ;)
So excited
Thanks

PR #245, opened 2019-11-01
Add run-ahead options

shantigilbert:
This will let you add up to 6 frames of run-ahead globally, per system, or per game, batocera.conf needs to be updated to handle 
* global.runahead
* global.secondinstance
* [system].runahead
* [system].secondinstance
* [game].runahead
* [game].secondinstance

### DIFF
--- a/es-app/src/guis/GuiMenu.cpp
+++ b/es-app/src/guis/GuiMenu.cpp
@@ -876,7 +876,23 @@ void GuiMenu::openGamesSettings_batocera()
 			SystemConf::getInstance()->saveSystemConf();
 		}
 	});
-	
+	// run-ahead
+	auto runahead_enabled = std::make_shared<OptionListComponent<std::string>>(mWindow, _("RUN-AHEAD FRAMES"));
+	runahead_enabled->add(_("NO"), "0", SystemConf::getInstance()->get("global.runahead") < "1");
+	runahead_enabled->add("1", "1", SystemConf::getInstance()->get("global.runahead") == "1");
+	runahead_enabled->add("2", "2", SystemConf::getInstance()->get("global.runahead") == "2");
+	runahead_enabled->add("3", "3", SystemConf::getInstance()->get("global.runahead") == "3");
+	runahead_enabled->add("4", "4", SystemConf::getInstance()->get("global.runahead") == "4");
+	runahead_enabled->add("5", "5", SystemConf::getInstance()->get("global.runahead") == "5");
+	runahead_enabled->add("6", "6", SystemConf::getInstance()->get("global.runahead") == "6");
+	s->addWithLabel(_("RUN-AHEAD FRAMES"), runahead_enabled);
+
+	// second instance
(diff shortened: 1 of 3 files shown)


Darknior:
WOW incredible !! Are you adding it to reduce latency in games ?
I only found this explanation : https://www.factornews.com/actualites/les-emulateurs-voient-l-avenir-43956.html
I don't really understand what we can configure ...
Thanks a lot

shantigilbert:
Yes runahead helps with lag, not all cores can use it but those that can really benefit from it.

Darknior:
Ok thanks, i have think it was already implemented in Retroarch ... i wish the wiki will be updated to explain how to use it :)

shantigilbert:
Yes this is just so that it can be set from emulationstation.

lbrpdx:
@shantigilbert but have you done the corresponding work in `configgen`? I can't see a PR for that yet.

shantigilbert:
@lbrpdx Unfortunately no, I was just getting ready to comment on that, I do not really use Batocera (only batocera-emulationstation), and I do not know Python that well, I use a custom bash script I made for my system.

Darknior:
Ok, it's cool to share this work with us :)
Thanks

lbrpdx:
@shantigilbert can you please share the part of your bash script where you take this addition into account, so that we can evaluate how much work it would be to do the corresponding modification in `configgen`?

shantigilbert:
Its actually really easy once you get the value from .runahead and .secondinstance (in this case its $2)

```
if [ "${2}" == "false" ] || [ "${2}" == "none" ] || [ "${2}" == "0" ]; then 
			echo 'run_ahead_enabled = "false"' >> ${RACONF}
			echo 'run_ahead_frames = "0"' >> ${RACONF}
		else
			echo 'run_ahead_enabled = "true"' >> ${RACONF}
			echo "run_ahead_frames = \"${2}\"" >> ${RACONF}
		fi
```

And for second instance
```
[ "${2}" == "1" ] && echo 'run_ahead_secondary_instance = "true"' >> ${RACONF} || echo 'run_ahead_secondary_instance = "false"'  >> ${RACONF} 
```

Sorry if I knew more Python or how Batocera really works I would have opened a PR with the changes.

nadenislamarre:
can it wait 5.25 ?

lbrpdx:
It’s not a lot of work, but this feature is quite CPU intensive, so I’d like to get some testing done on several architectures before committing this.

shantigilbert:
Yes it is CPU intensive, which is the reason I added it as per system/game as well, since this option in Retroarch is either on or off globally, with this option you can set it for some systems and off for others. I use it mostly for NES and SNES

nadenislamarre:
i'm not understood this one. is the configgen part already done ?

lbrpdx:
Not yet. It needs to be done. It’s not too hard to do, but I did not want to side track anything before 5.24 was out.

lbrpdx:
Configgen code PR: https://github.com/batocera-linux/batocera.linux/pull/1045

shantigilbert:
Ahh then it's all good :)

PR #258, opened 2019-11-13
Added "zfast" and "flatten-glow" shader sets.

lbrpdx:
Added the ES menu entries for https://github.com/batocera-linux/batocera.linux/pull/1049 (please merge the two PRs at the same time!).


nadenislamarre:
should we really translate these words ?!?

Darknior:
We already have "none, default, retro, scanlines, curvature, enhanced" 
Translated in "Aucun, default, retro, scanlines, courbé, amélioré" .... all is not translated.

- ZFast must not be tr in "ZRapide" ... but maybe "Nid d'abeille" to have a sense.
- flatter-glow can be tr in "Aplat & Lueur"

PR #263, opened 2019-11-19
Small fix on pt_BR emulationstation2.po

shantigilbert:
This prevented compilation


sabotendarsama:
Thanks for the correction in the translation I had submitted, @nadenislamarre you can accept that it's ok.

PR #271, opened 2019-11-24
New option to use d-pad as analog stick for NGC

lbrpdx:
For bartop/arcade cabinets where only a d-pad stick is available, enable this option to use it as the analog control stick on Gamecube. Many games can be playable with this option - which can be configured system-wide or on a game-by-game basis.

Needs the corresponding PR on configgen/dolphinControllers.py (which is being added at the same time)


nadenislamarre:
conflict

lbrpdx:
Useless now that https://github.com/batocera-linux/batocera.linux/pull/1066 is automatic.

PR #295, opened 2019-12-19
Allow gamelist subsets storage/customisations per system 

fabricecaruso:
- Created "subset" element to group includes & avoid repetitions of subset attributes in includes.
- Gamelist customisation per system is done using attribute appliesTo="viewname". 

```
<subset name="option" appliesTo="grid">
  <include name="yes">./samplefile0.xml</include>
  <include name="no">./samplefile1.xml</include>
</subset>
```


Darknior:
I love this subset option :)
I suppose you don't break the old system for compatibility ;)

And now we can select the viewname by system in ES ?
To have a different view by system ?
Nothing to write in theme himself ...
Thanks

Darknior:
And i release you also add "appliesTo="grid"" !!!
We can add now options by VIEW ? And this option will be hide if i don't use GRID view ?

PR #300, opened 2019-12-26
fix: screenscraper jaguar

jdorigao:
Signed-off-by: Juliano Dorigão <jdorigao@gmail.com>


fabricecaruso:
Hi, 

This could have been achieved differently :

- In es_systems.yml there is not "platorm:" value for jaguar. 
Adding "platform:   atarijaguar" in this file instead would have been a better way to do it. 
"platform" value is meant for that and should be in sync with PlatformId.cpp

- In ES, you don't have to remove old "string" values (these are retropie names), and you can add new ones using the same const value. ( see VIDEOPAC_ODYSSEY2 or SEGA_SG1000, I added multiple names declarations for that console. It's on purpose not to erase retropie names )

{ "atarijaguar",			ATARI_JAGUAR }, -> Doesn't have to be removed.
Just add the new one in the "// batocera specific names" section using the same const :  
{ "jaguar",			ATARI_JAGUAR } without removing the other one, and it will do the job.

jdorigao:
Nice!
Can I leave it to you to make this modification, so would it be the right way?

fabricecaruso:
Which one ??? ;-)

jdorigao:
The jaguar scraper requires {"jaguar", ATARI_JAGUAR}

fabricecaruso:
You have a second PR with it !?

PR #301, opened 2019-12-26
add: screenscraper jaguar

jdorigao:
Signed-off-by: Juliano Dorigão <jdorigao@gmail.com>


jdorigao:
@fabricecaruso Would that be the right form?

fabricecaruso:
Why did you open a second PR ? Is it a mistake ?

jdorigao:
To make the jaguar scraper you need a jaguar platform and not a atarijaguar

fabricecaruso:
As I said, this is because in "es_systems.yml" there is not "platform:" value for jaguar. It should be "platform: atarijaguar".
Anyway, we can add this one.

PR #332, opened 2020-01-20
Fix video path with random fallback

a-bertil:
Possibility to put a fixed video in the themes.
In the video tags indicate the path of the video in the <path> tag.
It is possible to use the following syntax: {random|path of the video}. If the video is not found, the videos of the platform are displayed randomly.

### DIFF
--- a/es-app/src/views/SystemView.cpp
+++ b/es-app/src/views/SystemView.cpp
@@ -267,8 +267,19 @@ void SystemView::populate()
 				if (extra->isKindOf<VideoComponent>())
 				{
 					auto elem = (*it)->getTheme()->getElement("system", extra->getTag(), "video");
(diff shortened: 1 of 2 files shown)


fabricecaruso:
Mmm Sorry, but I totally don't understand what you are trying to do.
You want to set a fixed video, and use {random} is the "fixed video" is not found ?

Well you can already do that in the theme, just duplicating <path> item like that :

<video>
   <path>{random}</path>
   <path>./myfixedvideo.mp4</path> 
</video>

Isn't that enough ?

a-bertil:
You're right, it does work.
I did the opposite, my fault. 
```
<video>
   <path>./myfixedvideo.mp4</path> 
   <path>{random}</path>
</video>
```
thank you

PR #346, opened 2020-01-25
add scrape for arcade systems

a-bertil:
added a function to the screenscraper scraper to be able to retrieve mame sub-systems (ex: cps1, cps2, sega classics...). Systems are created as collections. Do you think it's interesting to offer this feature ?


fabricecaruso:
I don't understand the logic : 

You want to create custom collections for arcade systems & automatically add games to the collection when you scrape them because you meet an arcade system id ?

Why do that during scrapping ?

Plus : "arcadesystemname" metadata does not have a declaration, so I don't understand how it can work.

I can understand the interest to store the arcade name in the metadatas. This info could also be added as a new filter. Then, in my POV, the best solution would have been to allow the creation of custom collection from filters.

a-bertil:
Yes, you get the idea.
I do this during the scraping because as you say the data is not available in the "arcadesystemname" metadata and it is possible to retrieve this information during the scraping.

fabricecaruso:
I'm sorry, but I really don't like the "use case" because you do that during scrapping :
A user with all its gamelists created, tranfering them to a fresh install can't create these collections on the new SD without scraping all again. 

Plus : As "arcadesystemname" is not declared, your code can't work fine ( see get/set declarations in metadata.cpp ).

For these reasons, I can't accept your PR now.

But, your PR starts from a very good idea. 
We just have to think of the best way to achieve that in a more generic "use case".

- Adding "arcadesystemname" metadata to gamelists & scrapping this information is a good idea ( not as an id but with the system name using the table yo encoded ) .
- Adding a new filter on this metadata is a good idea too.

The remaining question is how to create the custom collections & with easy access & not during scraping :

I have to two possible ideas :

1/ Either, make those collections "automatic collections" when "arcadesystemname" exists. 
Users will only have to "check" options to show/hide those collections.
The pro : if you scrap a new game or remove a game, it will automatically be added/removed. 
The cons : If we do that, we must ensure the "automatic collections" are not present in the list if no game has the corresponding "arcadesystemname" metadata to avoid "pollution".

2/ Add an menu option "create custom collection from filter". Users will have to do that manually and select the good filter themself. 
The pro : if you want to create a collection from all "mario" games, just input "mario" in "all games", select "create custom comllection from filter" and the "mario" collection will be created from this filter....
The con : It would be more complex for users, because they'll have to select the systems one by one in the filter list, and manually input the collection name. Plus, the collections won't be automatically updated.

Maybe we shoud do both ( but not for the same reasons )....

What's do you think about it ?

a-bertil:
the basic idea was to make something simple for the user the solution based on automatic collections seems to me more interesting for this use case. I'm going to work in this direction

The creation of collections from filters is also interesting but more for personalized collections.

a-bertil:
updated. The code needs to be improved but the principle is there.

fabricecaruso:
Well, I like it... but.... I still have some remarks & requests before merging ;-) :

- Your code crashes immediatly because you forgot to increment mDefaultGameMap & mGameTypeMap in metadata.cpp ( needs to be 22 ) -> Please fix this.

- We always see all subsystems in the collection list even if the collection is empty. 
It would be better not to show them when they are empty.
To achieve that you have to : 
         - Change the prototype for "inline std::map<std::string, CollectionSystemData>**&** getAutoCollectionSystems() { return mAutoCollectionSystemsData; };"  ( to add reference & )
         - make populateAutoCollection method public
         - populate the collection if needed in addSystemsToMenu like this :

```
std::map<std::string, CollectionSystemData>& autoSystems = CollectionSystemManager::get()->getAutoCollectionSystems();

	autoOptionList = std::make_shared< OptionListComponent<std::string> >(mWindow, _("SELECT COLLECTIONS"), true);

	// add Auto Systems

	for (std::map<std::string, CollectionSystemData>::iterator it = autoSystems.begin(); it != autoSystems.end(); it++)
	{
		if (it->second.decl.displayIfEmpty)
			autoOptionList->add(it->second.decl.longName, it->second.decl.name, it->second.isEnabled);
		else
		{
			if (!it->second.isPopulated)
			{
				CollectionSystemManager::get()->populateAutoCollection(&(it->second));				
				if (it->second.system->getRootFolder()->getChildren().size() == 0)
					continue;
			}

			autoOptionList->add(it->second.decl.longName, it->second.decl.name, it->second.isEnabled);
		}
	}

```

- What is the use to keep "ScrapeArcadeSystems" setting & the menu option ? 
It seems totally useless now that it is managed by collections. 
The Scraper should always scrape the information, and it should be transparent for users.
Don't you agree ?

a-bertil:
Done
for the last point I had kept the optional scrape because it was the case for the rateing. I think it's actually not very useful. I remove the setiing

fabricecaruso:
Mmmm everything seemed fine... 
Until I read the 3rd commit.... 

Why threadedScraper::processArcadeSystemCollection ????

Why clear the arcadesystemname metadata at that point ? Why clear it anyway ? Especially when there's a call to importScrappedMetadata next that will import & replace the value you just cleaned ? This code can't work correctly !

Why do that processing only in the threaded scrapper ? What about the manual scrapper ? The behaviour must be the same.

Why don't you test ARCADE in ScreenScraperRequest::processGame ?
But, Is there really a possibility we meet a "systeme.id" known in your screenscraper_arcadesystemid_map table  when it's not an ARCADE platform ?

Honesty, I don't understand why you added this !?

a-bertil:
I'm forced to remove it because we have to check the parent platform. It's mainly because of the neogeo. (there may be other systems in this case that I didn't add).
Where I added the code, this is one of the only times I can simply retrieve the platform being scraped.
During the scraping I only get the platform from the game.

To make it simple, for the neogeo that I scrap for the neo-geo or for arcade I would have the id of the neo-geo. So I have the meta "neogeo" for the arcade games and for the neogeo.

If I don't delete the information for the other platforms it will create duplicates in the collections.

I'd really prefer to do it in the specific part of screenscraper but I'd rather limit my modifications. I'll take another look at it.


PS : Oui

fabricecaruso:
Please contact me on batocera's discord server. It will be easier to finalize this.
https://discordapp.com/invite/ndyUKA5
My username is f.caruso

a-bertil:
done

PR #348, opened 2020-01-26
Wifi SSID enumeration/selection using "batocera-wifi" script

fabricecaruso:


lbrpdx:
Oh, that is SUPER COOL! Thank you!!

PR #368, opened 2020-02-10
fixed '--hash' option (depcreated for Linux?)

lbrpdx:
I left it for the Windows build, not sure if it's also still needed there (@fabricecaruso?).


nadenislamarre:
hum, we should analyze first. not sure it must be removed.
i've an item where f.caruso asked me to add the option in configgen.

lbrpdx:
I couldn’t find it in the current retroarch doc, nor in —help description. I didn’t look into the libretro code, but it looks deprecated, at least for Linux.

nadenislamarre:
ok, looks an old option. this is ok.

PR #373, opened 2020-02-12
Menus reorganisations 

fabricecaruso:
Using groups...
+ Support menuGroup in themes 
+ Better looking for missing bios screen


lbrpdx:
Many pages in the wiki will require updates.

PR #381, opened 2020-02-17
fix kodi inputs

nadenislamarre:


Darknior:
Thanks a lot

PR #451, opened 2020-04-18
Added additional formats for background music

lbrpdx:
Background music: added support for .flac, .wav and popular .mod chiptune formats (.mod, .xm., .stm, .s3m, .far, .it, .mtm, .669...)


lbrpdx:
#452 is the proper and clean PR for this

PR #461, opened 2020-04-26
user and system disk usage

nadenislamarre:


fabricecaruso:
Win32ApiSystem version are to be fixed... (homepath can be relative / system path can be another letter ). But I'll manage it in a future PR.

PR #480, opened 2020-05-10
Fix : GridView compatibility with retropie was broken with some themes.

fabricecaruso:
+ White background when custom backgroundImage was defined and backgroundCornerSize was set at 0 on tiles 
+ clipping issue on Retropie gridviews using autolayout.


nadenislamarre:
should i update 5.26 rc1 with it ? or we can wait 5.27 ?

PR #482, opened 2020-05-12
Add "skip to next song" menu item

lbrpdx:


shantigilbert:
This is awesome! Wonder if you could map to a button?

lbrpdx:
There is no button left... or none that can be mapped to a simple SNES controller.

shantigilbert:
But many other people use controllers that have a few buttons left. :)

lbrpdx:
This is a (much) larger question: make the buttons configurable in ES.

shantigilbert:
Yeah I was talking something more like in `ViewController.cpp` 

```
	// next song
	if(config->isMappedTo("l3", input) && input.value != 0) // batocera
	{
		// next song
		AudioManager::getInstance()->playRandomMusic(false);
		return true;
	}
```
since l3 is not a button used on ES (not sure if it works) 

But I agree a configurable button would be harder.

fabricecaruso:
- Why did you rename the group "volume" to "control" ?
- Is it really the best place to do that ? It requires at least 7 moves to change the song... Would not it be better in the "openQuitMenu_batocera_static" menu, near "LAUNCH SCREENSAVER" function ?
- L3/R3 button are free...

lbrpdx:
* I renamed it "controls" because "skip to next song" is not related to volume any more, and I thought this would be the most natural placement for the feature. If I remove and put it in the "Quit" menu, I'll put "Volume" back obviously.
* Good call on the "Quit" menu... only "select" + one selection = 2 clicks. But same question here, I think the menu should be renamed. All the more that there already is a screensaver option in there. Any suggestion to rename this "Quit" menu?
* Mmm... Not sure we want to waste a button for this feature. Again, I think we'd prefer to have a full configuration, which goes well beyond the scope of this PR. But I can add it, as you say, we still have 2 buttons left (not on my arcade controller, but OK).

fabricecaruso:
> Mmm... Not sure we want to waste a button for this feature. 

Or you can use : right analog right that may be more intuitive.... 
Right analog directions are not used at all

lbrpdx:
I already pushed a commit with L3. The problem with right analog is that it lets you think you could potentially go back in the playlist.. which is not possible with the current code (and I don't think it's worth investing time on this, Emulation is not a full-fledged music player by any means).

fabricecaruso:
@nadenislamarre Should we merge for the 5.26 release or shall we wait for it to be merged for 5.27 ?

lbrpdx:
No response on this after 2 days - @nadenislamarre I hope all is fine with you.
Can we merge it and see later if you want update ES's git revision in Batocera's 5.26 branch?

nadenislamarre:
it is ok for me for 5.26.

PR #503, opened 2020-06-03
add: pc88 to ScreenScraper

jdorigao:


es-app/src/LangParser.cpp
```diff
@@ -56,7 +56,7 @@ void LangInfo::extractLang(std::string val)
 		{ { "kr", "korea", "k" }, "kr", "kr" },
 		{ { "cn", "china", "hong", "kong", "ch", "hk", "as", "tw" }, "cn", "cn" },
 
-		{ { "in", "�ndia" }, "in", "in" },
+		{ { "in", "�ndia" }, "in", "in" },
```
fabricecaruso, line 59: Is this change normal ?

jdorigao: This change is only appearing here on github

PR #507, opened 2020-06-06
fix br flag

azvarel:


Darknior:
We must not add it, we must rename it from BZ to BR and i think add PT flag because i miss it

coiotato:
PT flag will be added too? :)

Darknior:
i don't know, i stand @fabricecaruso ask me if i must draw some new flags. For the moment PT flag is not created because i never found any PT game, only BR games.

fabricecaruso:
As discussed on discord, this renaming will be included in my next PR, so I close this PR.

PR #548, opened 2020-07-20
update locale fr_FR

evoflash72:


evoflash72:
Wrong PR, updated in #551

PR #583, opened 2020-09-01
TextListComponent/ImageGridComponent : Add themable scrollbar indicator

fabricecaruso:
Using new theme properties on imagegrid & textlist components : scrollbarColor, scrollbarCorner and scrollbarSize

es-theme-carbon needs to be bumped.

image


nadenislamarre:
if possibe i would like to merge that one after 5.27 release, now, we target bug fixing only (and translations)

PR #601, opened 2020-10-01
Big Fix to ES 5.27.2

seeiis:


nadenislamarre:
did you regenareted the .po ? you should start from the correct po. you get too many modifications then.

seeiis:
ok let me check the file, i made a lot of modifications making more clean spanish translate...

PR #632, opened 2020-11-01
Fix .keys files paths to include original game extension

fabricecaruso:
+ use batocera-systems with --filter parameter for better performance


nadenislamarre:
cmd += " --filter " + system; you should add " " around system, even if it is always a keyword without a space

PR #645, opened 2020-11-03
pt-PT translation

coiotato:


nadenislamarre:
arg conflict because i updated it ;-(

PR #669, opened 2020-11-17
Added missing translation (es)

danfossi:
Added missing translation


meegle84:
The commit message is confusing, because is no for "es" translation, is for "ex_MX"

PR #671, opened 2020-11-17
Don't upgrade all packages unless requested (refresh only)

lbrpdx:
UI for https://github.com/batocera-linux/batocera.linux/pull/2489 with a new button "UPDATE ALL" separate from "REFRESH".


fabricecaruso:
Fine to me

PR #708, opened 2020-12-08
Enable RGA scaling option in ES (for OGA screen)

lbrpdx:
See a screenshot of this libretro option in the corresponding batocera.linux PR.

### DIFF
--- a/es-app/src/SystemData.cpp
+++ b/es-app/src/SystemData.cpp
@@ -384,6 +384,7 @@ EmulatorFeatures::Features EmulatorFeatures::parseFeatures(const std::string fea
 		if (trim == "colorization") ret = ret | EmulatorFeatures::Features::colorization;		
 		if (trim == "padtokeyboard") ret = ret | EmulatorFeatures::Features::padTokeyboard;		
 		if (trim == "joystick2pad") ret = ret | EmulatorFeatures::Features::padTokeyboard;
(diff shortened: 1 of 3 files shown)


fabricecaruso:
I don't agree with this PR. It should be a custom feature instead.

lbrpdx:
It has to be something you can set at the global, system and game level. Pretty much like we do for "pixel perfect" and "smooth". They're all at the same level on retroarch's options, and should all be following the same pattern.

shantigilbert:
But isn't rga scaling only available for Rockchip handheld devices?

lbrpdx:
@shantigilbert you made a great point. I tested only on OGA, but you are right, I can't find the same option on x86_64.

shantigilbert:
Yep this option is only enabled with librga and librga only works on Rockchip as far as I know, so it might work on other Rockchip devices but I have not tested it.

fabricecaruso:
> It has to be something you can set at the global, system and game level.
System and game levels already exist. Global level does not.
> this option is only enabled with librga and librga
If it has to be filtered, extending es_features is the good way. The associated script can be extended to filter elements the same way es_systems scripts does.

lbrpdx:
We need it at the global level (it makes sense for OGA, and might even be something we could put on by default).
I agree with everything you said, extending es_features is the way to go.

lbrpdx:
Can be replaced with the new capacities of es_features.cfg, closing it.

PR #749, opened 2021-01-30
Use GPU vertex shader to compute final mvpMatrix

rtissera:


rtissera:
@fabricecaruso this is replacing the CPU based matrix-multiplication in Renderer by GPU-based one using shader uniforms.

As the Transform4x4f implementation is pure C without SIMD, this should reduce CPU usage on ARM boards.
The cost for a GPU-based 4x4 matrix multiplication is really low compared to CPU-based similar operation.

rtissera:
Closing this as setProjection is not even used anymore, we can do better.

PR #804, opened 2021-03-16
Add timezone selection and 12-hour mode for clock

lbrpdx:


fabricecaruso:
I merged it but :
- Does not compile so I commited & fix.

After seeing it i'm wondering this :
- there's a "show clock" option in UI options & a "show clock in 21-h fmt" in advanced options.
Wouldn't it be better to merge everything in a single option :
"SHOW CLOCK" "OFF / 24 HOURS FORMAT / 12 HOURS FORMAT" in the UI options !?

EDIT  : I commited also a fix for Windows because %p doesn't work on windows with french locale.

lbrpdx:
What did not compile originally? I assume it was under Windows?

For your question, I thought about it. But to me, the 12h-format is more a "localization" thing that should live next to the language and timezone settings. "Show clock" is more a "screen personalization" option. 
Longer term, there are other parts in Batocera where we should also adapt the 24/12-hour time format (movie recordings and screenshots for example... they have a non-standard form for naming the files with the hour hardcoded in).
But I'd be OK with "off / 24-hour format / 12-hour format" too if you prefer it this way.

es-app/src/ApiSystem.cpp
```diff
@@ -1433,3 +1433,38 @@ std::vector<std::string> ApiSystem::getRetroachievementsSoundsList()
...
+
+	std::vector<std::string> folderList = { "/usr/share/zoneinfo/" };
+	for (auto folder : folderList)
+	{
+		for (auto continent : Utils::FileSystem::getDirContent(folder, false))
+		{
+			for (auto file : Utils::FileSystem::getDirContent(continent, false))
+			{
+				std::string short_continent = continent.substr(continent.find_last_of('/') + 1, -1);
+				if (short_continent != "posix" and short_continent != "right")
```
fabricecaruso, line 1451: That line didn't compile ( it's not python... and does not exist )

lbrpdx: Wait... what? Logical "and" is the C++ equivalent of good old C "&&" that cool kids use nowadays https://en.wikipedia.org/wiki/Operators_in_C_and_C%2B%2B#C.2B.2B_operator_synonyms

And "and" compiles fine with gcc 9.3.0 on Linux. Isn't that supported with your environment?

PR #814, opened 2021-03-29
Trimmed down the list of timezones (deprecated out)

lbrpdx:


lbrpdx:
Can I get some eyes on this, please? It reduces the list by about 80 deprecated timezones, which makes the UI more manageable.

PR #856, opened 2021-05-04
Clean up content downloader cached files (system -> developer)

lbrpdx:
Otherwise you can end up with GB of useless files.


lbrpdx:
Replaced by #862

PR #866, opened 2021-05-08
MaxVRAM=256 for Odroid XU4

jdorigao:


nadenislamarre:
@fabricecaruso for MaxVRAM, maybe we could add a CMake definition to inject from es.mk ?

PR #867, opened 2021-05-10
Better timezone configuration

lbrpdx:
If no timezone has been defined, try to autodetect the best timezone from the IP address, and otherwise default to "Europe/Paris" (default TZ until now). Requires scripts from https://github.com/batocera-linux/batocera.linux/pull/3882

### DIFF
--- a/es-app/src/ApiSystem.cpp
+++ b/es-app/src/ApiSystem.cpp
@@ -1475,11 +1475,24 @@ std::vector<std::string> ApiSystem::getTimezones()
 	return ret;
 }
 
+std::string ApiSystem::getCurrentTimezone()
+{
+	LOG(LogInfo) << "ApiSystem::getCurrentTimezone";
+	auto cmd = executeEnumerationScript("batocera-timezone get");
+	std::string tz = Utils::String::join(cmd, "");
+	remove_if(tz.begin(), tz.end(), isspace);
+	if (tz.empty()) {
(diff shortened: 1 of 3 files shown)


lbrpdx:
If you look at `/etc/init.d/S26system`, the source of truth for timezone configuration is `batocera.conf`, which generates /etc/timezone at boot time (through `batocera-config`). So I think the current implementation is the best. If there is nothing in batocera.conf, it is autodetected. If there is something (and it's a valid TZ) this is what is selected.

nadenislamarre:
can you confirm that in case of no network, the call is not blocking (curl https://ipapi.co/json/) ?

fabricecaruso:
Why don't you use a compact timezone list, like this one ?
https://github.com/filipdanic/compact-timezone-list/blob/master/index.js#L293

lbrpdx:
> Why don't you use a compact timezone list, like this one ?
> https://github.com/filipdanic/compact-timezone-list/blob/master/index.js#L293

Because we already have the list of supported timezones elsewhere on the system. You can shrink it down again with a compact list if you think that's useful. 
But don't forget that it's a list that is used once by users, and auto-populated at first launch through `batocera-timezone detect`.

lbrpdx:
> can you confirm that in case of no network, the call is not blocking (curl https://ipapi.co/json/) ?

Confirmed, it returns immediately.

PR #890, opened 2021-05-26
Update HU

igyenes:
Update HU


nadenislamarre:
thanks a lot for this update.
there are some untranslated messages remaining, do you need help on them ?

hu    686 translated messages, 97 untranslated messages.

PR #906, opened 2021-06-06
Themes : Add support for ifArch/ifNotArch attributes -> Allow filtering elements/subsets by board to optimize slow boards

fabricecaruso:
NOTE : Please bump the theme, I added ifArch/ifNotArch attributes to deactivate some things for slow boards.

+ Add ES Video renderer information 
+ better organisation of information screen
+ Splash : Fix crash when gamesplash.xml is invalid 
+ Win32 only / Retrobat : if es-update script last line is a zip file, uncompress it using internal ES installer

### DIFF
--- a/es-app/src/Win32ApiSystem.cpp
+++ b/es-app/src/Win32ApiSystem.cpp
@@ -812,6 +812,94 @@ static std::string getScriptPath(const std::string& name)
 	return "";
 }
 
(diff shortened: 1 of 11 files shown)


nadenislamarre:
std::string getArchString()
i could define you the string, it would be easier.

PR #937, opened 2021-07-23
quit from thread should not call exit

nadenislamarre:
otherwise... when restarting pw, it crashes


crcerror:
ah missed that, thanks

PR #941, opened 2021-07-30
Add a button on ES web services to kill running emulator

lbrpdx:


nadenislamarre:
2 questions :
- should we use batocera-es-swissknife in programs ? i allowed it just to "help", but for me, this script is not nice to use in a robust solution, just for end users for some shortcut. (however, i can undrestand the need, so why not...)
- we should find a way to extend es instead of writting function not related to es in es. for example, we could do a flask proxy that calls es when es function, and python calls otherwise (but it is one another service to start). maybe f.caruso will have a better idea.

lbrpdx:
1. I think using `batocera-es-swissknife` is fine... maybe change its name for something a little more 'neutral' like `batocera-cli` or whatever, but the tools in the swiss knife are really useful and I prefer reuse that than reinventing the wheel.
2. Yes, I'd love @fabricecaruso to chime in. Talked to the Retrobat guys, it looks like the web services are not used under Windows, so there might be a way to do better on the longer term.

PR #969, opened 2021-08-31
Do not group systems for PER SYSTEM ADVANCED CONFIGURATION

NostalgicDavid:
Fixes Nintendo Entertainment System and Sega Megadrive not being shown in "GAMES SETTINGS" - "PER SYSTEM ADVANCED CONFIGURATION" menu.

Nintendo Entertainment System has the same group as Famicom Disk System and was hidden in "PER SYSTEM ADVANCED CONFIGURATION" menu. Settings for Famicom Disk System had no impact on Nintendo Entertainment System games because different setting prefixes in batocera.conf

### DIFF
--- a/es-app/src/FileData.cpp
+++ b/es-app/src/FileData.cpp
@@ -776,7 +776,7 @@ const std::vector<FileData*> FolderData::getChildrenListToDisplay()
 	auto sys = CollectionSystemManager::get()->getSystemToView(mSystem);
 
 	std::vector<std::string> hiddenExts;
-	if (!mSystem->isGroupSystem() && !mSystem->isCollection())
+	if (mSystem->isGameSystem() && !mSystem->isCollection())
 		for (auto ext : Utils::String::split(Settings::getInstance()->getString(mSystem->getName() + ".HiddenExt"), ';'))	
 			hiddenExts.push_back("." + Utils::String::toLower(ext));
 	
@@ -979,7 +979,7 @@ std::vector<FileData*> FolderData::getFilesRecursive(unsigned int typeMask, bool
 	SystemData* pSystem = (system != nullptr ? system : mSystem);
 
 	std::vector<std::string> hiddenExts;
-	if (!pSystem->isGroupSystem() && !pSystem->isCollection())
+	if (pSystem->isGameSystem() && !pSystem->isCollection())
 		for (auto ext : Utils::String::split(Settings::getInstance()->getString(pSystem->getName() + ".HiddenExt"), ';'))
 			hiddenExts.push_back("." + Utils::String::toLower(ext));
 
--- a/es-app/src/FileFilterIndex.cpp
+++ b/es-app/src/FileFilterIndex.cpp
@@ -912,7 +912,7 @@ bool CollectionFilter::createFromSystem(const std::string name, SystemData* syst
 
 	mSystemFilter.clear();
 
-	if (system != nullptr && !system->isCollection() && !system->isGroupSystem() && system->getName() != "all")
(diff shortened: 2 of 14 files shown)


fabricecaruso:
That's not a good fix for now. 

I wasn't informed the 'nes' systems was grouped to itself in the recent es_systems -> The feature was never indented to group system to themselves, so, yes, you're right, this will require to do some fixes.

But :
- Removing isGroupSystem test is wrong. If you do that, you're gonna show every pure "group" systems like "amiga" or "msx" which is wrong : Those system don't have advanced configurations.
- Instead of removing "system->isGroupSystem()", we should replace this call with "!system->isGameSystem()". System grouped to themself will be visible, & pure groups will continue to be hidden.
- There's a lot of other places in the code to manage. Every call to isGroupSystem should be checked & modified that way. 

Can you do that, please ?

NostalgicDavid:
@fabricecaruso I have uploaded a new PR. Please review it carefully, because I don't know what majority of changed code do!

Can you also take a look at the places that I did not change?
https://github.com/batocera-linux/batocera-emulationstation/blob/master/es-app/src/views/SystemView.cpp#L787
https://github.com/batocera-linux/batocera-emulationstation/blob/master/es-app/src/guis/GuiGameOptions.cpp#L228
https://github.com/batocera-linux/batocera-emulationstation/blob/master/es-app/src/services/HttpApi.cpp#L59 I don't know where the value gets used
https://github.com/batocera-linux/batocera-emulationstation/blob/master/es-app/src/SystemData.h several places
https://github.com/batocera-linux/batocera-emulationstation/blob/master/es-app/src/SystemData.cpp I changed some places, others remain unchanged

Also I have found a regression. With grouping enabled FDS games are shown in Favorites collection two times. It should be very easy for you to find/fix.

fabricecaruso:
What are those 27 commits ? 
Your branch is " 27 commits ahead, 1 commit behind batocera-linux:master. "

Please rebase your branch on batocera-linux:master, then I'll merge it once done.

NostalgicDavid:
I have updated the PR.
:heavy_check_mark: Rebased.
:heavy_check_mark: Fixed regression. FDS games are properly shown in Favorites collection.
:heavy_check_mark: Removed change in [ISimpleGameListView.cpp#L202](https://github.com/batocera-linux/batocera-emulationstation/blob/master/es-app/src/views/gamelist/ISimpleGameListView.cpp#L202) because subexpression would always be true.

New unchanged files:
https://github.com/batocera-linux/batocera-emulationstation/blob/master/es-app/src/CollectionSystemManager.cpp#L967 caused regression: FDS games twice displayed in Favorites and Last Played
https://github.com/batocera-linux/batocera-emulationstation/blob/master/es-app/src/views/gamelist/ISimpleGameListView.cpp#L202 subexpression would always be true

Other unchanged or partially unchanged files:
https://github.com/batocera-linux/batocera-emulationstation/blob/master/es-app/src/views/SystemView.cpp#L787 expression would always be false
https://github.com/batocera-linux/batocera-emulationstation/blob/master/es-app/src/guis/GuiGameOptions.cpp#L228 subexpression would always be true
https://github.com/batocera-linux/batocera-emulationstation/blob/master/es-app/src/services/HttpApi.cpp#L59 probably no changes needed, but I don't know where the value gets used
https://github.com/batocera-linux/batocera-emulationstation/blob/master/es-app/src/SystemData.h probably no changes needed, isGroupSystem() and mIsGroupSystem in several places
https://github.com/batocera-linux/batocera-emulationstation/blob/master/es-app/src/SystemData.cpp isGroupSystem() and mIsGroupSystem in several places, I changed some places, others left unchanged

PR #970, opened 2021-08-31
GuiFileBrowser: Add SELECT button for better UI

NostalgicDavid:
Makes file browser GUI more intuitive. In other places actions are confirmed with A button which is needed for entering folders in file browser.
SELECT button can help confused users.

### DIFF
--- a/es-app/src/guis/GuiFileBrowser.cpp
+++ b/es-app/src/guis/GuiFileBrowser.cpp
@@ -38,6 +38,22 @@ GuiFileBrowser::GuiFileBrowser(Window* window, const std::string startPath, cons
 
 	if (mOkCallback != nullptr)
 	{
+		mMenu.addButton(_("SELECT"), "select", [&]
+		{
+			auto path = mMenu.getSelected();
+
+			if (mTypes == FileTypes::DIRECTORY && path.empty())
+			{
+				mOkCallback(mCurrentPath);
+				delete this;
+			}
+			else if (!path.empty() && (mTypes == FileTypes::DIRECTORY || !Utils::FileSystem::isDirectory(path)))
+			{
+				mOkCallback(path);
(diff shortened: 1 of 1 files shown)


fabricecaruso:
I'm divided.

When you set the focus to the SELECT button you can't see anymore what item is selected in the list. 
Doing that, when you press the button, you can't confirm with an eye what you SELECT.
So : In my point of view, it's less intuitive than having no button at all.

Pressing A is generally enough -> It works like that in every other menu.

Addin the button would be a good thing if the selected item was still visible, but there's no "always show selection" option in ComponentList, and there's no management in themes to differenciate selection when the list has the focus, or not (ie make the selection line gray when the focus is lost).

My personal opinion is that it's more intuitive without the button.

NostalgicDavid:
With this PR the [start] button on gamepad will work as previously.

I personally never thought about pressing [start] button. I tried A, B, X and Y buttons and afterwards concluded that folder selection for screensaver pictures/videos does not work at all. I edited emulationstation config file manually.

I got the knowledge about [start] button only from code when I tried to "fix the bug".

fabricecaruso:
What is the relationship between the gamepad [start] button and the select button in the UI ?

fabricecaruso:
What ??? It's not a question about renaming the button or not.

It's a question to determine if it's pertinent to add this button, knowing that you can't see the selected item in the list when the button has the focus.

In my optinion : it's not pertinent, because it's not intuitive at all.

NostalgicDavid:
@fabricecaruso Sorry! I have misunderstood this question:
> What is the relationship between the gamepad [start] button and the select button in the UI ?

Both ([start] button and SELECT button in UI) run exactly the same code and do exactly the same thing. Each user can decide which method is more comfortable.

PR #972, opened 2021-09-01
minor grammar correction

Hew-ux:


nadenislamarre:
where there are 2 symbol in a ending symbol (like ? ! :) the rule is to have a space before AND after.

lbrpdx:
@nadenislamarre That's a syntax rule in French, but not in English ;) 
This PR is correct from an English grammar perspective.

Hew-ux:
What lbrpdx said. English has punctuation always attached to the last character in a written sentence, it looks awkward (from an English-perspective) to have a space.

nadenislamarre:
hey man, you've already the default language as english, you want the default punctuation too ;-)
more seriously, while all is done with the double space style, i suggest we keep it.
an other point is that we should move english correction after v32 while it is already hard to get up to date translation, if we change the english now, for sure, some translations will be missing.

Hew-ux:
alright, that's ok. I'll turn this into a draft and we'll merge this after v32.

PR #986, opened 2021-09-18
Add ui setting to ignore leading articles when sorting

not43s:
* This PR adds a new UI Setting option for ignoring leading articles when sorting.
* "The Legend of Zelda" is now sorted like "Legend of Zelda", "A Boy and His Blob" -> "Boy and His Blob", etc.
* I set the default value to `false` to maintain existing behavior.
* ~~The articles default to "a,an,the", but can be adjusted in `batocera.conf` if the user wishes.~~
* The articles can be defined per language

Thanks!

### DIFF
--- a/es-app/src/FileSorts.cpp
+++ b/es-app/src/FileSorts.cpp
@@ -1,3 +1,4 @@
+#include <Settings.h>
 #include "FileSorts.h"
 
 #include "utils/StringUtil.h"
@@ -74,10 +75,39 @@ namespace FileSorts
 	bool compareName(const FileData* file1, const FileData* file2)
 	{
 		if (file1->getType() != file2->getType())
+		{
 			return file1->getType() == FOLDER;
+		}
+		// we compare the actual metadata name, as collection files have the system appended which messes up the order
+		auto name1 = ((FileData *) file1)->getName();
+		auto name2 = ((FileData *) file2)->getName();
+		const bool ignoreArticles = Settings::getInstance()->getBool("IgnoreLeadingArticles");
+		if (ignoreArticles)
+		{
+			const auto articles = Utils::String::commaStringToVector(_("A,AN,THE"));
+			name1 = stripLeadingArticle(name1, articles);
+			name2 = stripLeadingArticle(name2, articles);
+		}
+		return Utils::String::compareIgnoreCase(name1, name2) < 0;
+	}
 
-		// we compare the actual metadata name, as collection files have the system appended which messes up the order		
-		return Utils::String::compareIgnoreCase(((FileData*)file1)->getName(), ((FileData*)file2)->getName()) < 0;
+	std::string stripLeadingArticle(const std::string &string, const std::vector<std::string> &articles)
+	{
+		const auto candidate = Utils::String::trim(string);
+		const auto index = candidate.find_first_of(" \t\r\n");
+		if (index == std::string::npos)
+		{
+			return string;
+		}
+		const auto maybeArticle = candidate.substr(0, index);
+		for (auto &article: articles)
+		{
+			if (Utils::String::compareIgnoreCase(article, maybeArticle) == 0)
+			{
+				return Utils::String::trim(candidate.substr(article.length()));
+			}
+		}
+		return string;
 	}
 
 	bool compareRating(const FileData* file1, const FileData* file2)
--- a/es-app/src/FileSorts.h
+++ b/es-app/src/FileSorts.h
@@ -81,5 +81,7 @@ namespace FileSorts
 
 	bool compareSystemReleaseYear(const FileData* file1, const FileData* file2);
 	bool compareReleaseYearSystem(const FileData* file1, const FileData* file2);
+
+	std::string stripLeadingArticle(const std::string &string, const std::vector<std::string> &articles);
 };
 #endif // ES_APP_FILE_SORTS_H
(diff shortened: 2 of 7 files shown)


nadenislamarre:
i think that "a,an,the" should be a translatable string

Darknior:
Excellent idea, for now i've always do it myself in my gamelist but this option will be better :)
For sure @nadenislamarre has a good idea too, because when i have read your PR i have think i will must use the batocera.conf to add the french articles .... if there are already translated like all other ES texts it will be better :)

not43s:
> i think that "a,an,the" should be a translatable string

Done and squashed.

fabricecaruso:
I think it's not a good idea to sort names without the articles, but continue display things with them.

Imagine such an order :
- LEGO star wars
- The Legend of Zelda
- Luigis mansion

What does a T does between a L and a L !? 
It's absolutely not intuitive at all.

You can't alter sorting without also altering the display, or it's not understandable by common users.

In my point of view : **It should not be an option for sorting. Instead, It could be an option when scrapping names**
 -> "The Legend of Zelda" or "Legend of Zelda (The)" -> That's the good way to go.

not43s:
> What does a T does between a L and a L !?
> It's absolutely not intuitive at all.

I'm honestly surprised to hear this. This [is a very common thing](https://www.google.com/search?hl=en&q=sorting%20ignore%20article) to do with English titles. There is also prior existing precedent for this in emulation OSes whether it seems intuitive or not e.g. [retropie has this feature](https://github.com/RetroPie/EmulationStation/pull/729) and it behaves the same way albeit English only.

> You can't alter sorting without also altering the display, or it's not understandable by common users.

This is how its supposed to work though and it is intuitive for those who wish to use it. It is also the reason this option defaults to **disabled** to maintain existing behavior and _not_ confuse common users.

> In my point of view : **It should not be an option for sorting. Instead, It could be an option when scrapping names**
> -> "The Legend of Zelda" or "Legend of Zelda (The)" -> That's the good way to go.

You can certainly do this; I was naming my own games similarly by hand to work around this issue e.g. _Legend of Zelda, The: Ocarina of Time_. I suppose you just find the first non-alphanumeric character (colon, dash, etc.) and do the rearranging before it (I can't think of a title off the top of my head that doesn't adhere to this convention).

But this approach also has issues:

- You have to rescrape all your games when you want to change the behavior, while my approach is just a toggle switch.
- It permanently alters the canonical name of the title; now _anywhere_ the title appears the article is adjusted, which certainly doesn't seem right in my eyes. My approach keeps the canonical name.
- At least in American English, this is not the normal convention when ignoring articles while mine is.

Now there is a different approach to tackling this problem that probably satisfies both our desires, but it is more invasive code-wise and so I didn't want to start with it (although I was thinking about adding it).

This solution is to add a new `sortName` attribute to the `game` element in _gamelist.xml_ and has implications:

- Requires a rescrape to change behavior which has downsides
- Allows the same sorting behavior
- Keeps the canonical name
- Makes sorting even more useful e.g. _Final Fantasy VII_ can be renamed to _Final Fantasy 7_ so it is sorted below _Final Fantasy IX_, and likewise for other titles with unique names
- There needs to be a way to edit `sortName` manually now in case scraper gets it wrong or the user wants something different
- There is prior precedent for this; [retropie also has this feature](https://github.com/RetroPie/EmulationStation/pull/240)
- This approach requires changes to `gamelist.xml` schema which is not trivial and must be done carefully and so the overall coding effort is greater.

The rationale for this feature in lieu of this more robust solution was that it is pretty non-invasive and I did not expect it to be controversial. I am certainly open to a different approach but I can't say I'm a fan of altering the canonical name.

fabricecaruso:
I'm French, so, it would never have occurred to me to withdraw articles for sorting.

" The One " is different to me than just " One ". Articles sometimes have a meaning in the sentence, so why sort it without articles ? Especially if the article is still normal on display. 
Do that in French would shock me a lot... It would make it really hard to find what you're looking for.

But I'm French, (American) English is not my native language.

After some research, I found that Wikipedia says : https://en.wikipedia.org/wiki/English_articles#Effect_on_alphabetical_order
So maybe you're right & it's a common practice in America. @lbrpdx Can you confirm it's a common practice !?

I personnaly don't like retropie approach & prefer yours.

lbrpdx:
I am with @not43s on this. And I am not sure why you see that as a problem in French?

Would you sort « La Légende de Zelda » with all others « La »? It might not be a great example because both names start with L. But I think I’d expect to mentally skip the article. Another example: I’d search for « J » to play « Le Jour de la Tentacule ».

But I admit I’m biased, all my scraping is done in English, I don’t play the French versions.

fabricecaruso:
"La Légende de Zelda" c'est 'L' et 'L' donc... mauvais exemple.

Mais "L'aigle d'or" c'est L, pas A... le trier en A me choque clairement.  
"La ferme se rebelle" L, pas F -> L'article fait partie intégrante du titre. Le nom du truc c'est pas "Ferme se rebelle", mais bien "LA ferme se rebelle". 
Et voir "L'aigle d'or" en plein milieu d'élements qui commencent tous par A, quand l'affichage conserve l'article, je trouve ca juste illisible.

Apres chacun fait ce qu'il veut... Tant que c'est une option désactivée par défaut.

not43s:
@fabricecaruso How do you feel about the `sortName` solution?

fabricecaruso:
Bad ! As I said : I personnaly don't like retropie approach & prefer yours.

PR #995, opened 2021-09-30
Feature : "LAUNCH THIS GAME AT STARTUP" option 

fabricecaruso:


nadenislamarre:
is it safe to save launch command ?
SystemConf::getInstance()->set("global.bootgame.cmd", bootgame->getState() ? fileData->getlaunchCommand(LaunchGameOptions(), false) : "");
instead of system/game ?
i fear it can break after an upgrade.

PR #971, opened 2021-09-01
Spelling, grammar and clarity pass-over

Hew-ux:
Disambiguated some of the more confusing terminology, corrected grammar where appropriate.

PR #1004, opened 2021-10-05
Spelling, grammar and clarity pass-over 2

Hew-ux:
Edits the English strings in both the .cpp and .po files to use correct spelling/grammer and/or be more informative as to their function.

Sorry translators, I've tried my best to carry over the strings that have been translated by affected lines but there may be some instances where a translated string has been blanked out/no longer appropriate. Here's a list of the most notable changes:
[notable strings.txt](https://github.com/batocera-linux/batocera-emulationstation/files/7282541/notable.strings.txt)

This is a successor to my previous (closed) PR: https://github.com/batocera-linux/batocera-emulationstation/pull/971


Hew-ux:
I looked into that and unfortunately found out that various aspects of EmulationStation actually do display text according to their capitalisation (like the pop-up notification, error messages, etc.). ie. the casing carries meaning. It is unfortunate for languages that don't use or require capitalisation but this seems like it's necessary at least for the base English strings.

If it were up to me we wouldn't be directly referencing the English strings but instead use code IDs for translation proposed but we have to work with what we've got here.

With that said I did search for common English similies (like OK, Okay, etc.) and didn't find any egregious use of them.

evoflash72:
Try with this:
cat locale/lang/en/LC_MESSAGES/emulationstation2.po | grep msgid | tr a-z A-Z | sort | uniq -c | sort -n

Hew-ux:
My friend I think you are misunderstanding.

Try running the command you have just told me to run, all should become more clear.

The English strings aren't an available translation, _they're a part of the code_.

PR #1024, opened 2021-10-27
Fix : getFileSize returns wrong size for >= 8Gb files

fabricecaruso:
So they are…hashed before scrapping -> time to way is almost eternal....

### DIFF
--- a/es-app/src/scrapers/ScreenScraper.cpp
+++ b/es-app/src/scrapers/ScreenScraper.cpp
@@ -238,7 +238,7 @@ void ScreenScraperScraper::generateRequests(const ScraperSearchParams& params,
 		path += "&romtype=rom";
 
 		std::string fileNameToHash = params.game->getFullPath();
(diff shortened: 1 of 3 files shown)


fabricecaruso:
Yes. See wikipedia ;-) https://fr.wikipedia.org/wiki/Types_de_donn%C3%A9e_du_langage_C

nadenislamarre:
i don't find it in the wikipedia page you point to

nadenislamarre:
unsigned long long yes, but no unsigned unsigned long long

nadenislamarre:
[ 60%] Building CXX object es-core/CMakeFiles/es-core.dir/src/utils/FileSystemUtil.cpp.o
/home/nicolas/dev/batocera-emulationstation/es-core/src/utils/FileSystemUtil.cpp:1222:12: error: duplicate ‘unsigned’
 1222 |   unsigned unsigned long long getFileSize(const std::string& _path)

nadenislamarre:
gcc linux n'aime pas non plus

PR #1036, opened 2021-11-11
Update preload labels in dev options

Hew-ux:
PRELOAD UI > PRELOAD UI ELEMENTS (Reduce overall menu lag at the cost of storage space)
PRELOAD MEDIAS FILESYSTEM (REDUCE UI LAGS OVER STARTUP TIME) > PRELOAD METADATA MEDIA (Reduce gamelist lag at the cost of storage space)

### DIFF
--- a/es-app/src/guis/GuiMenu.cpp
+++ b/es-app/src/guis/GuiMenu.cpp
@@ -1081,13 +1081,13 @@ void GuiMenu::openDeveloperSettings()
 	// preload UI
 	auto preloadUI = std::make_shared<SwitchComponent>(mWindow);
 	preloadUI->setState(Settings::getInstance()->getBool("PreloadUI"));
-	s->addWithLabel(_("PRELOAD UI"), preloadUI);
+	s->addWithDescription(_("PRELOAD UI ELEMENTS"), _("Reduce overall menu lag at the cost of storage space"), preloadUI);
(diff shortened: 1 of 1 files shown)


nadenislamarre:
i think that all strings are uppercase (of almost all) in es

Hew-ux:
Most are. Deceptive labels have not been (you know, the smaller text underneath the main label.) Eg. any option in advanced per system settings, the "swap confirm and cancel buttons" option (in that same dev many mind you), etc.

Either this all caps is changed to regular casing, or all the other hundred or so descriptive labels are changed to all caps. I think the former is quicker.

fabricecaruso:
@Hew-ux @nadenislamarre 

"at the cost of storage space" ???? Seriously !?
That's **absolutely not** what happens and the label you changed/added are now totally wrong !!!

"PRELOAD UI" Preloads the gamelist UIs ( instead of doing a lazy loading ) -> Either it takes more time @ startup. Either it introduce lags 1st time you enter a gamelist.

"PRELOAD METADATA MEDIA" -> Either it takes more time @ startup checkin' if media files exist. Either medias files existance is checked only when required which can produce lags when browsing the UI.

Please fix the labels.

Hew-ux:
Noted, I'll make a new PR that fixes these labels.

PR #1037, opened 2021-11-11
Update translation fr_FR

evoflash72:


nadenislamarre:
msgstr "ÉFFACER LES CACHES"
=> there is no É before 2 FF.

PR #1040, opened 2021-11-17
correct preload labels

Hew-ux:
As noted by @fabricecaruso in #1036.
Draft until confirmed that the strings still fit in correctly.


fabricecaruso:
Please add a mention to explain that if the option is checked, **ES takes more time to start/boot** 
I originally added the description to mention that -> Either it reduce some lags, either it takes more time to boot.

Hew-ux:
Integrated your suggestions, labels are now perfectly clear as to what they do.
Strings fit in fine, ready to merge.

PR #1041, opened 2021-11-17
pt-PT translation

coiotato:


Hew-ux:
This PR has broken ES compilation due to non-escaped double-quotes.

In line 1363, `"SAVES"` should be `\"SAVES\"` or `"""SAVES"""`.

PR #1042, opened 2021-11-19
fixes saveslot selection, brightness typo, autosave slot name

dhwz:
Just fixing a few things I've already fixed in our fork.

### DIFF
--- a/es-app/src/ApiSystem.cpp
+++ b/es-app/src/ApiSystem.cpp
@@ -967,7 +967,7 @@ const char* BACKLIGHT_BRIGHTNESS_NAME = "/sys/class/backlight/backlight/brightne
 const char* BACKLIGHT_BRIGHTNESS_MAX_NAME = "/sys/class/backlight/backlight/max_brightness";
 #define BACKLIGHT_BUFFER_SIZE 127
 
-bool ApiSystem::getBrighness(int& value)
+bool ApiSystem::getBrightness(int& value)
 {
 #if WIN32
 	return false;
@@ -1011,7 +1011,7 @@ bool ApiSystem::getBrighness(int& value)
 #endif
 }
 
-void ApiSystem::setBrighness(int value)
+void ApiSystem::setBrightness(int value)
 {
 #if !WIN32	
 	if (value < 5)
@@ -1049,7 +1049,7 @@ void ApiSystem::setBrighness(int value)
 
 	count = write(fd, buffer, strlen(buffer));
 	if (count < 0)
-		LOG(LogError) << "ApiSystem::setBrighness failed";
+		LOG(LogError) << "ApiSystem::setBrightness failed";
 
 	close(fd);
 #endif
--- a/es-app/src/ApiSystem.h
+++ b/es-app/src/ApiSystem.h
@@ -204,8 +204,8 @@ public:
 
 	void callBatoceraPreGameListsHook();
 
-	bool	getBrighness(int& value);
-	void	setBrighness(int value);
+	bool	getBrightness(int& value);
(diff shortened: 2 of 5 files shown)


dhwz:
@fabricecaruso to what change do you not agree?

fabricecaruso:
It's highlighted ! -> this change -> if (incrementalSaveStates && nextSlot > 0)

dhwz:
But why it's definitely ignoring the incremental setting right now?
We already had users which complained because if auto save/load is set to off it is incrementing the slot with every launch no matter if incrementalsavestates is enabled or not.
There are users which want to stay on slot 0.

fabricecaruso:
> "There are users which want to stay on slot 0."

Why run a new game on slot 0 if there's already an existing savestate on slot 0 ? 
In this case : you'll always overwrite the current save in slot 0 !
Really !? 
In my optinion, it's very problematic to do that, because you overwrite something that was not intended to be overwritten !!!!

Running a new game = use a new free slot.
**You should NEVER overwrite an existing save**, unless "incremental savestate" is unchecked & **ONLY if you start the game using an existing savestate** from an existing slot !

Doing that is also a good way to create new slots. Unless you change the slot id in retroarch, with your piece of code, you'll always be with slot 0, and you cannot create new slots anymore.

> "It is incrementing the slot with every launch"

It's absolutely not true : The used slot is the next free slot... If you don't save to a new slot, the number won't change. If you have a  saves in 0, the next free slot is 1... So not "with every launch"

I personnally don't like this change at all & don't agree to merge it.

Hew-ux:
Is it possible that this is related to this issue? https://github.com/batocera-linux/batocera.linux/issues/4738 I remember @nadenislamarre said he was looking to fix it but I'm not sure if this is going to conflict with that or if this is the fix itself.
Also related: https://github.com/batocera-linux/batocera.linux/issues/4631

dhwz:
@fabricecaruso I understand your arguments but you prevent giving users a choice how to use it.

There are a lot of users which just want to use one single save slot, for whatever reason we have not to decide why they do so.

Adding incrementalSaveStates doesn't break anything it just gives a choice to the user.
incrementalSaveStates = on -> the slot is incrementing if you start a new game
incrementalSaveStates = off -> the slot is not incrementing if you start a new game and stays on 0 (this is useful for any user which doesn't want to have multiple save slots and also doesn't want to use the save state manager).

Just think about it again, there are users which DON'T use the savestate manager at all, and all these users get forced to a new slot and have then to switch back to 0 in RetroArch so they can load the slot 0 again or save to it if needed.

If you don't want to merge it fine.

fabricecaruso:
And what about users ( probably majority ) in this use case : 
I have my own personal saves on slot 0. 
My child runs a new game and he performs a save... 
My save in slot 0 is overwritten. 

With your code : I loose everything ( and I am very angry ! ).
With the current code : As he started a new game, it's a new slot, so my child's save would have been in slot 1. Mine is still in 0 -> This is the behaviour I want to keep, because I don't want accidental overwrites.

Then, maybe, "incremental savestate" option should be split in 3 instead : "YES", "NO - NEW GAME USE NEW SLOT" (current code), "NO - ALWAYS USE SLOT 0" (your way to do)

dhwz:
Ok yes that would be an alternative

fabricecaruso:
@dhwz I just commited a split of the savestate options to include the parameter we discussed in this PR : https://github.com/batocera-linux/batocera-emulationstation/pull/1078
Then, I'm closing your PR.

dhwz:
@fabricecaruso  Thanks!

PR #1058, opened 2021-11-28
Integrate PixelArcade support

Hew-ux:
User needs to separately install all the necessary dependencies and make sure everything else is configured correctly in Batocera itself.

This change just makes it so that they can use batocera-emulationstation with it as well as the custom fork of emulationstation PixelArcade has. One fork that natively supports PixelArcade: https://github.com/alinke/PIXEL/tree/wip/EmulationStation-kai

Would like to have another dev look over this first before I switch this off of draft.

To help with organising my thoughts I made notes in draft copies containing just the code that was needed to be used and where it should go:
[modified.zip](https://github.com/batocera-linux/batocera-emulationstation/files/7612702/modified.zip)

### DIFF
--- a/es-app/src/SystemScreenSaver.cpp
+++ b/es-app/src/SystemScreenSaver.cpp
@@ -22,6 +22,7 @@
 #include "SystemConf.h"
 #include "ImageIO.h"
 #include "utils/Randomizer.h"
+#include "HttpReq.h"
 
 #define FADE_TIME 			500
 
@@ -124,6 +125,13 @@ void SystemScreenSaver::startScreenSaver()
 
 			PowerSaver::runningScreenSaver(true);
 			mTimer = 0;
+
+			if (mCurrentGame != NULL) {
+				HttpReq req("http://127.0.0.1:8080/arcade/stream/" + HttpReq::urlEncode(mCurrentGame->getSystem()->getName()) + "/" + HttpReq::urlEncode(mCurrentGame->getFileName()) + "?t=" + HttpReq::urlEncode(mCurrentGame->getName()));
+				if(req.status() != 200){
+         			//std::cout "http://127.0.0.1:8080/arcade/stream/" << mCurrentGame->getSystem()->getName() << "/" << mCurrentGame->getFileName() << "failed.";
+           		}
+			}
 			return;
 		}
 	}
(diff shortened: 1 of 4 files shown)


fabricecaruso:
For your information : I'll **NEVER** merge that kind of specific calls into ES.

Don't forget that ES is also used by Retrobat ( Windows ), Emulelec, RetroLX. 
They don't have such a service. 

The code you commit there must run on Windows and must be importable in the other forks without consequences.

Use & extend scripting instead.

n-i-x:
> For your information : I'll **NEVER** merge that kind of specific calls into ES.
> 
> Don't forget that ES is also used by Retrobat ( Windows ), Emulelec, RetroLX. They don't have such a service.
> 
> The code you commit there must run on Windows and must be importable in the other forks without consequences.
> 
> Use & extend scripting instead.

Are there specific platform related bits you are referring to? The device which this adds support for will work on any platform. The device is https://pixelcade.org and is an LED and LCD marquee and has a Java based API server which this PR intends to call. Any device which has a JRE is capable of supporting the Pixelcade. For example, the pixelcade devs said Batocera wasn't compatible and I got it working by using an embedded JRE and starting the API server. The only missing piece is the ES side to call the API when a game/system is selected.

The alternative implementation is to extend the scripting functionality in batocera to call an external script from these same locations. If you're okay with that many exec calls in ES, that will indeed make the device work as intended.

If support for the device is added directly to ES, there should be a config option and a UI option to enable support which will enable these API calls. They shouldn't be made if the device isn't listening. It really just depends on if Batocera wants to call out support for Pixelcade and have it built-in or if the scripting engine should be used for these types of integrations. I'm kinda biased since I have the device in my cabinet (I'm not associated with pixelcade in any way) and having an option to enable/disable the pixelcade would be nice, but I can make the device work using either method the maintainers deem appropriate and can work with you to make it happen.

n-i-x:
FYI the full API is documented here: https://pixelcade.org/api/

fabricecaruso:
The only acceptable ways to do this in ES are :
- Either use & extend scriting with Scripting::fireEvent(...). In that case, it's handled by scripts located in /usr/share/emulationstation. If the scripts are missing, then, nothing happens.
- Either use an external script/exe ( batocera-pixelarcade ? ), invoked with ApiSystem ( see how it's done with batocera-config, batocera-bluetooth, batocera-sync... in ApiSystem.cpp ). External scripts presence can be detected ( see how it's done in ApiSystem::isScriptingSupported ) & then functionnality is disabled when the scripts does not exist.

Hew-ux:
I agree with that, it would be better just to extend the scripting functionality. This was just too see if it could work in the first place.
I'll close this PR and see if there's a more uniform approach to adding the (optional) API calls. I don't think this needs to be Batocera-specific either, so I'll investigate making it an option saved solely within ES.
If you're not comfortable with that though I can look into making it a package within just Batocera too; there are so many ways this can be achieved.
Thanks for the feedback.

PR #1062, opened 2021-11-30
add challenge indicator option to retroachievements section

Hew-ux:
complements https://github.com/batocera-linux/batocera.linux/pull/5156

### DIFF
--- a/es-app/src/guis/GuiMenu.cpp
+++ b/es-app/src/guis/GuiMenu.cpp
@@ -1852,6 +1852,13 @@ void GuiMenu::openRetroachievementsSettings()
 	retroachievements_screenshot_enabled->setState(SystemConf::getInstance()->getBool("global.retroachievements.screenshot"));
 	retroachievements->addWithLabel(_("AUTOMATIC SCREENSHOT"), retroachievements_screenshot_enabled);
 	retroachievements->addSaveFunc([retroachievements_screenshot_enabled] { SystemConf::getInstance()->setBool("global.retroachievements.screenshot", retroachievements_screenshot_enabled->getState()); });
+
+	// retroachievements_challenge_indicator
+	auto retroachievements_challenge_indicators = std::make_shared<SwitchComponent>(mWindow);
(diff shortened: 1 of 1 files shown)


fabricecaruso:
@nadenislamarre  Ok for me. You can merge this one as soon as the other PR is merged into batocera-linux

nadenislamarre:
but why don't we use the feature file ?

Hew-ux:
> but why don't we use the feature file ?

I'm not sure what you mean, this is just a toggle option in RetroArch (when RetroAchievements are enabled).

Darknior:
Yes it is the same solution we already using for other Retroachievement options, i think it is the good one.

nadenislamarre:
there is a es_features.yml file to declare features. why to not use it.
for existing es lr feature, we must move them to the file too.

Hew-ux:
I don't think we're on the same page here. Any system that supports RetroAchievements will have challenge indicators on by default. This is a toggle to turn that off, just like with other settings related to RetroAchievements like leaderboards, auto-screenshot, etc.

If you'd like to see an example, refer to https://github.com/batocera-linux/batocera.linux/pull/5156 as mentioned in the first post. This is something you would see in any system that supports RetroAchievements when RetroAchievements are activated.

nadenislamarre:
@Hew-ux options for es should not be defined in es while there are specific to batocera,
they should be defined here : https://github.com/batocera-linux/batocera.linux/blob/master/package/batocera/emulationstation/batocera-es-system/es_features.yml

Darknior:
@nadenislamarre Retroachievement options have never been write in es_features ?
Only in the "libretroConfig.py " file ... 
I speak for Python code. Libretro options about Retroachievements are not in es_features for sure.
There we only tag "cheevos" or not, no more

Hew-ux:
I'm pretty confused now. @nadenislamarre if you could show me the code implementation of what you are suggesting I'd probably understand it better.

fabricecaruso:
@nadenislamarre Retroachievement screen is not a game/system option screen, so it's not dependant on es_features. 
There's no way we can implement this with es_features for now.

As the other PR is merged. I merge this one.

PR #1063, opened 2021-12-01
Add resize option to decorations submenu + minor fixes

Hew-ux:
Allows the user to opt to use the original resolution + ratio of the image they have selected as the tattoo. "AUTO" behaviour is to resize to 225 px proportionately (similar to current live behaviour).
Also adds contingencies if the user tattoo is too large for the screen
and consistent terminology to some other items in the decorations submenu (so as not to confuse the term "overlay" (which is what RetroArch refers to bezels as sometimes) with "bezel").

Complements https://github.com/batocera-linux/batocera.linux/pull/5161

### DIFF
--- a/es-app/src/guis/GuiMenu.cpp
+++ b/es-app/src/guis/GuiMenu.cpp
@@ -2156,14 +2156,14 @@ void GuiMenu::openGamesSettings_batocera()
 						});
 
 				// tattoo and controller overlays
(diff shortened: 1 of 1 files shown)


nadenislamarre:
not a fan of all these hard coded values. it means it will fails in other resolutions. why these options are not in es_features.yml ?

nadenislamarre:
note that tatoo is now added on the hud, so, next beta x86_64, attention must be taken for standalones too.

PR #1065, opened 2021-12-03
rk3326: cleaning odroidgoa

jdorigao:
odroidgoa is part of the rk3326 group

### DIFF
--- a/es-app/src/guis/GuiMenu.cpp
+++ b/es-app/src/guis/GuiMenu.cpp
@@ -1550,7 +1550,7 @@ void GuiMenu::openSystemSettings_batocera()
 
 	std::shared_ptr<OptionListComponent<std::string>> overclock_choice;
 
-#if ODROIDGOA || GAMEFORCE || RK3326
+#if GAMEFORCE || RK3326
 	// multimedia keys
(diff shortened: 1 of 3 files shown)


fabricecaruso:
**ODROIDGOA** does not exist anymore in Batocera !?
Anyway even if it does not exist anymore in Batocera, you can leave the #define it's not a big deal.

What about other forks @rtissera @shantigilbert. 
Do you add/use a definition for ODROIDGOA in your forks ?

shantigilbert:
Seems weird to me, is this basically just changing the name from ODROIDGOA to RK3326? Since GAMEFORCE is still there that is also a RK3326 device, I personally use both definitions as the Gameforce has its own small tweaks, but I would assume this is done to support all the other RK3326 devices (that I don't), but like the OGA and the GameForce some of them are not the same (e.g. screen size) 

I have no issue with this change, but to me its seems unnecessary.

jdorigao:
Okay, so I didn't pay attention to the forks, sorry!

https://github.com/batocera-linux/batocera.linux/pull/5160

- RK3326 with support for: 
    - Anbernic RG351P
    - Odroid Go Advance
    - Odroid Go Super
    - Powkiddy RGB10
    - RK2020
    
- Game Force with support for:
    - Golden Game Force
    
Note: Although GameForce is rk3326 it is separated due to some particular differences.

PR #1073, opened 2021-12-11
add fireEvents for Pixelcade

Hew-ux:
Trying this again: https://github.com/batocera-linux/batocera-emulationstation/pull/1058

But this time, using script calls.

### DIFF
--- a/es-app/src/views/gamelist/ISimpleGameListView.cpp
+++ b/es-app/src/views/gamelist/ISimpleGameListView.cpp
@@ -312,6 +312,14 @@ bool ISimpleGameListView::input(InputConfig* config, Input input)
 		return true;
 	}
 
+	// Pixelcade
+	FileData* cursor = getCursor();
+	SystemData* system = getSystem();
+	if (system != NULL)
+	{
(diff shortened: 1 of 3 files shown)


fabricecaruso:
- Event will not fire in many cases. See the calls to FileData::speak for better locations ( BasicGameListView / DetailedContainer ).
- "cursorhover" name is not really appropriate. "game-selected" would be a better name ( following convention of game-start and game-end events ).

Hew-ux:
So if the name "game-selected" is used, then the bash script would have to wait for the "gameSelected" event, right? Because for the "game-start" event it waits for the "gameStart" event. I'm not sure why they aren't the same, actually. Is there a file that translates the event names or something? I couldn't find any if so.

Also, this is a work in progress. I've only implemented it in one location, and am waiting on the ones with the hardware to test that it works first before writing the code in similar style to all other relevant places (ie. the places in https://github.com/batocera-linux/batocera-emulationstation/pull/1058).

Hew-ux:
Nice work on your PR at https://github.com/batocera-linux/batocera-emulationstation/pull/1075! That basically does everything I was intending to do with this PR, so I'll close this one to avoid redundancy.

PR #1074, opened 2021-12-12
Fix an issue with non-scraped games on Linux

lbrpdx:
emulatorlauncher.py on Batocera always requires an argument for ` -gameinfoxml` (even if the file doesn't exist).

### DIFF
--- a/es-app/src/FileData.cpp
+++ b/es-app/src/FileData.cpp
@@ -501,8 +501,13 @@ std::string FileData::getlaunchCommand(LaunchGameOptions& options, bool includeC
 		command = Utils::String::replace(command, "%GAMEINFOXML%", Utils::FileSystem::getEscapedPath(fileInfo));
 	else
 	{
+#ifdef WIN32
 		command = Utils::String::replace(command, "%GAMEINFOXML%", "");
 		Utils::FileSystem::removeFile(fileInfo);
+#else
+		// Linux emulatorlauncher.py (Batocera) always requires a /tmp/game.xml (even if non existent file)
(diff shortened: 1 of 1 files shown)


nadenislamarre:
so why not to make it optionnal ?
hum https://github.com/batocera-linux/batocera.linux/blob/master/package/batocera/core/batocera-configgen/configgen/configgen/emulatorlauncher.py#L612
required is false.
what is the error message ?

lbrpdx:
This line says that the argument “-gameinfoxml” is optional but when it is there, it requires a string behind it. 
You can’t just put “-gameinfoxml” without any string for the xml path, otherwise you’ll have an error message in es_launch_stderr.log telling that a missing parameter is required (and the game isn’t launched, back to ES).

fabricecaruso:
I don't like this fix at all...
- Either you have a problem when parsing commandlines : -gameinfoxml requires a 2nd parameter which is missing. -> You can add a fix to ignore missing argument when the next one startswith -
- Either, the "-gameinfoxml" argument need to be totally set by ES. ( like this `command = Utils::String::replace(command, "%GAMEINFOXML%", "-gameinfoxml " + Utils::FileSystem::getEscapedPath(fileInfo));` ). In this case we need to remove "-gameinfoxml" from the es_systems file

lbrpdx:
Thank you @nadenislamarre and @fabricecaruso for your comments. I replaced this with a fix in emulatorlauncher.py (https://github.com/batocera-linux/batocera.linux/pull/5238)

PR #1092, opened 2021-12-24
Refactored Renderer apis to use IRenderer interface. 

fabricecaruso:
This allow multiple drivers ( for Windows mainly ), and set the Renderer to use with es_settings.cfg file

For Windows, you can add the following entries in es_settings. Renderers are :
`<string name="Renderer" value="OPENGL 2.1 / GLSL" />` 
-> default, used if "Renderer" is not defined

or

`<string name="Renderer" value="OPENGL 2.1" />` 
-> forces the old OpenGL driver


netonjm (approved):

netonjm:
thanks for this :-)

PR #1110, opened 2022-01-14
add translation line

Hew-ux:
translation line for the new set introduced with one of the latest batocera betas


nadenislamarre:
not understood. you add one line ? github is showing several lines modified.

Hew-ux:
That's right.
The other lines are extra tabbing so it still lines up. The lines themselves are unmodified.

PR #1116, opened 2022-01-19
Track music history to stop songs replaying

wizzomafizzo:
I've changed the behaviour of the music player so it will keep a limited history of previously played songs in memory, and skip playing songs if they're in that list. So it will avoid situations where the current randomness is "not random enough" and repeats recent songs. I have had situations where songs have repeated many times and it gets annoying.

This solution isn't perfect but I think it does an adequate job of fixing this problem, can be easily removed later and adds very little overhead to the current system. It also works when per system music is enabled.

The history size is hardcoded, and I've just made a decision for it to be 40% of the size of the total song collection.

### DIFF
--- a/es-core/src/AudioManager.cpp
+++ b/es-core/src/AudioManager.cpp
@@ -17,6 +17,10 @@
 #include <unistd.h>
 #endif
 
+// batocera
+// Size of last played music history as a percentage of file total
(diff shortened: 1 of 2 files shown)


wizzomafizzo:
Oops I've messed this pull request up now. Will recreate with its own branch

PR #1118, opened 2022-01-20
Limit song title length sent to notification

wizzomafizzo:
If an mp3 file has an id3 v1 song title at or over 30 characters the notification window will overflow into the next id3 fields trying to display it.

The files I was using actually also have v2 tags, the default behaviour of that app I used to set them was to do both. It might be better to prefer v2 tags if both exist.


fabricecaruso:
I don't understand the problem. It displays fine when it's > 30 chars.

fabricecaruso:
Ok I'm stupid sorry. char title[30] for idv3 v1...

PR #1119, opened 2022-01-20
Add system menu option to view controller overlay

wizzomafizzo:
This adds a menu option to the system menu for viewing its controller overlay image. I would rather not hardcode the path but I couldn't see where/if this path is defined.

It's not ready to go in until high res overlay images have been included in the image (works fine but looks like crap)

### DIFF
--- a/es-app/src/guis/GuiGamelistOptions.cpp
+++ b/es-app/src/guis/GuiGamelistOptions.cpp
@@ -303,15 +303,6 @@ GuiGamelistOptions::GuiGamelistOptions(Window* window, IGameListView* gamelist, 
 			{
 				mMenu.addGroup(_("OPTIONS"));
 
-				std::string controllerOverlayPath = "/usr/share/batocera/controller-overlays/" + mSystem->getName() + ".png";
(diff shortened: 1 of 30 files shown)


fabricecaruso:
For your information : 
As EmulationStation has to be portable ( Paths must exist also on Windows, with EmuElec, RetroLX, etc... )
I'm currently refactoring the whole "paths thing" to have only one cpp file storing every "known" paths. 
You can see my work branch for my next PR including this code there :
https://github.com/fabricecaruso/batocera-emulationstation/blob/win/es-core/src/Paths.cpp#L20

In consequence, I won't accept new PRs with code using specific paths like this ( /usr/share/batocera/controller-overlays ), unless declared outside this class.

Maybe wait for my next PR to finish yours...

wizzomafizzo:
Oh that is a much better idea. Yes happy to wait for that to go in and then update this later

fabricecaruso:
Will you finish this PR ?

wizzomafizzo:
No plans to. Feel free to close if nobody else wants to implement

PR #1155, opened 2022-03-21
Shorten "REBOOT SYSTEM TO APPLY THE UPDATE" string

Hew-ux:
As it is right now, the "UPDATE IS READY" notification is too small to contain this string:
screenshot-2022 03 21-17h38 24

 This shortens it to fit.


nadenislamarre:
what about making more space, while this is not only for this message, and for all languages ?

Hew-ux:
I mean yeah, that would be preferable. This is just easier and has no consequences to do.

fabricecaruso:
Width is hardcoded for the notifications, based for 30 chars. It's on purpose to keep them aligned when there are multiple notifications, or when the text changes -> It avoids widths change.
So : I'm fine with this label change.

PR #1182, opened 2022-05-10
Add new option "SHOW EMPTY SYSTEMS"

fabricecaruso:
Will be useful for themers


Darknior:
So cool, thanks a lot, it will change my life for theming <3

pajarorrojo:
Thanks! a great step to keep themes up to date :)

PR #1183, opened 2022-05-11
Support Non-ASCII rom path in windows platform

dreampet:
Convert system command  from UTF-8 to ANSI code page to support Non-ASCII rom path in windows platform
before:
2022-05-11 19:23:23.800 [INFO]    "E:\RetroBat\emulationstation\emulatorLauncher.exe"  -p1index 0 -p1guid 030000005e0400008e02000000007200 -p1name "Xbox 360 Controller" -p1nbbuttons 11 -p1nbhats 1 -p1nbaxes 6  -system psp -emulator ppsspp -core  -rom "E:\RetroBat\roms\psp\\**G_古惑狼泰�?cso**"
2022-05-11 19:23:23.858 [ERROR]   rom does not exist

after:
2022-05-11 19:51:54.371 [INFO]    "E:\RetroBat\emulationstation\emulatorLauncher.exe"  -p1index 0 -p1guid 030000005e0400008e02000000007200 -p1name "Xbox 360 Controller" -p1nbbuttons 11 -p1nbhats 1 -p1nbaxes 6  -system psp -emulator ppsspp -core  -rom "E:\RetroBat\roms\psp\\**G_古惑狼泰坦.cso**"

### DIFF
--- a/es-core/src/platform.cpp
+++ b/es-core/src/platform.cpp
@@ -57,10 +57,10 @@ int runSystemCommand(const std::string& cmd_utf8, const std::string& name, Windo
 
 #define BUFFER_SIZE 8192
 
-	TCHAR szEnvPath[BUFFER_SIZE];
-	DWORD dwLen = ExpandEnvironmentStringsA(command.c_str(), szEnvPath, BUFFER_SIZE);
+	WCHAR szEnvPath[BUFFER_SIZE];
+	DWORD dwLen = ExpandEnvironmentStringsW(Utils::String::convertToWideString(command).c_str(), szEnvPath, BUFFER_SIZE);
 	if (dwLen > 0 && dwLen < BUFFER_SIZE)
-		command = std::string(szEnvPath);
+		command = Utils::String::convertFromWideString(szEnvPath);
 
 	std::string exe;
 	std::string args;


fabricecaruso:
I tested your PR with many files containing Non-ASCII chars ( also tried with the filename you provided ).

The code you wrote **is not working for me.** on any file containing Non-ASCII chars.
Also : It works fine with the current code.

Why did you use CP_ACP ?

dreampet:
@fabricecaruso Hi, I'm using the Simplified-Chinese version of Windows 10, in this version the current code only works fine with the option "Beta: Use Unicode UTF-8 for worldwide language support" checked.

You can see this option by going to Settings and then: All Settings -> Time & Language -> Language -> "Administrative Language Settings"

But after this option is checked, many other Non-UTF8 programs can not work fine.

It means that: The current code only works fine for the computer in which the option "Unicode UTF-8 support" is checked, but this PR makes many other computers in the default setting.

dreampet:
> Why did you use CP_ACP ?

@fabricecaruso maybe I should try to find a way to get the current codepage of the system instead of using hardcode CP_ACP.

fabricecaruso:
I don't understand what is wrong on your system with the current code... It's working fine here

dreampet:
@fabricecaruso  This is not the issue of my system, is a general issue on many other Asian versions of Windows systems, like Japanese, Korea, and Chinese.
I had found an API to get the current code page of the system, and only do the CP_ACP convert on the Non-UTF8 system, I'll push the new code later.

fabricecaruso:
It's not working either. 
GetACP() returns 1252 which is not CP_UTF8 ( GetACP probably never returns CP_UTF8 ), so it's still converting to CP_ACP, which does not work on my System.

It should work as it is, because we use ShellExecuteExW, not ShellExecuteExA and it's using multibyte chars.

What your codepage ? Here's a list of codepages : https://docs.microsoft.com/en-us/windows/win32/intl/code-page-identifiers

If there is a general problem with some codepage, maybe you need to limit this code to an hardcoded list of codepages ( 50220 to 50939 ? I don't know... )

dreampet:
@fabricecaruso  My system codepage is 936, and I can reproduce this problem in codepages 932, 949, and 950. 

by the way, I found an interesting thing yesterday:
This problem only occurs when the path contains odd Non-ASCII characters (like  `G_古惑狼泰坦.cso` ), If the ROM path contains even Non-ASCII characters (like `G_古惑狼泰坦坦.cso` ), the current code works fine, 

finally, I found the root cause is the function `ExpandEnvironmentStringsA`,  everything goes wrong after this function call.

I'll keep on this issue and make a more general path without a hardcode codepage list.

dreampet:
Maybe we shoud convert the command to `std::wstring` at first, and then use `ExpandEnvironmentStringsW` instead of  `ExpandEnvironmentStringsA`, I will test it later

fabricecaruso:
Interesting...
Yes, there's a mix in the code with A & W apis.
We'd better normalize all this to only use std::wstring, then use ExpandEnvironmentStringsW + write a new SplitPath use a widechar version.

rekcah1986:
Great! Someone has finally solved this problem. Thank you!

fabricecaruso:
Are you going to finish this PR ? 

The only necessary change is probably only that :

```
	WCHAR szEnvPath[BUFFER_SIZE];
	DWORD dwLen = ExpandEnvironmentStringsW(Utils::String::convertToWideString(command).c_str(), szEnvPath, BUFFER_SIZE);
	if (dwLen > 0 && dwLen < BUFFER_SIZE)
		command = Utils::String::convertFromWideString(szEnvPath);
```

Will you test this & change your PR ?

dreampet:
@fabricecaruso Hi, I have changed the code like you,  but I'm busy at work these days, tomorrow we have a three-day holiday, and I will test and update this PR ASAP.

fabricecaruso:
Perfect, thx

rekcah1986:
哥们太给力了，给你点赞~ @dreampet

PR #1192, opened 2022-05-20
Sony Analog Trigger Fix

TVsIan:
Implements a modified version of RetroPie's controller config. The root cause of the issue is that Sony controllers (DS3 and later) will report both a button press and an axis+ for the triggers. Emulationstation and most emulator configs will detect the button over the axis, and be unable to properly configure analog triggers.

* If the controller has at least 6 axis inputs, and the input being assigned is L2 or R2, ignore button input so the axis can be detected.
* If the button is held for a second, and there has been no axis input, assume the button is correct and assign it.

The fallback shouldn't likely be used except in the case of weird controllers or the user setting different buttons to L2/R2.


dmanlfc:
@TVsIan can you squash the commit but also try using...

`if(config->getDeviceNbAxes() >= 6)`

TVsIan:
I can try that. I squished it as much as I could, but I may just re-do it
as a new commit if that change works.

TVsIan:
Closed due to updated/cleaner version.

PR #1194, opened 2022-05-24
Trigger Input Filter v2

TVsIan:
This replaces the previous PR - cleaner commit history and axis count check.

### DIFF
--- a/es-core/src/guis/GuiInputConfig.cpp
+++ b/es-core/src/guis/GuiInputConfig.cpp
@@ -159,7 +159,7 @@ GuiInputConfig::GuiInputConfig(Window* window, InputConfig* target, bool reconfi
 
 
 			// filter for input quirks specific to Sony DualShock 3
-			if(filterTrigger(input, config))
+			if(filterTrigger(input, config, i))
 				return false;
 
 			// we are configuring, the button is unpressed or the axis is relaxed
@@ -376,7 +376,52 @@ void GuiInputConfig::clearAssignment(int inputId)
 	mTargetConfig->unmapInput(GUI_INPUT_CONFIG_LIST[inputId].name);
 }
 
-bool GuiInputConfig::filterTrigger(Input input, InputConfig* config)
+bool GuiInputConfig::filterTrigger(Input input, InputConfig* config, int inputId)
 {
+#if defined(__linux__)
+	// on Linux, some gamepads return both an analog axis and a digital button for the trigger;
+	// we want the analog axis only, so this function removes the button press event
+
+	if(config->getDeviceNbAxes() >= 6)
+	{
+		// digital triggers are unwanted
+		if(input.type == TYPE_BUTTON
+		&& (GUI_INPUT_CONFIG_LIST[inputId].name == "l2"
+		|| GUI_INPUT_CONFIG_LIST[inputId].name == "r2"))
+		{
+			if(mHoldingInput && mHeldTime < 1000)
+			{
+				return false;
+			}
+			else
+			{
+				return true;
+			}
+		}
+	}
+
+	// ignore negative pole for axes 2/5 only when triggers are being configured
+	if(input.type == TYPE_AXIS)
+	{
+		if(GUI_INPUT_CONFIG_LIST[inputId].name == "l2"
+		|| GUI_INPUT_CONFIG_LIST[inputId].name == "r2")
+		{
+			if(input.value == 1)
+				mSkipAxis = true;
+			else if(input.value == -1)
+				return true;
+		}
+		else if(mSkipAxis)
+		{
+			mSkipAxis = false;
+			return true;
+		}
+	}
+#else
+	(void)input;
+	(void)config;
+	(void)inputId;
+#endif
+
 	return false;
 }
--- a/es-core/src/guis/GuiInputConfig.cpp
+++ b/es-core/src/guis/GuiInputConfig.cpp
@@ -157,9 +157,11 @@ GuiInputConfig::GuiInputConfig(Window* window, InputConfig* target, bool reconfi
 				return false;
 			}
 
+			if (mHoldingInput)
+				mAllInputs.push_back(input);
 
 			// filter for input quirks specific to Sony DualShock 3
-			if(filterTrigger(input, config, i))
+			if(filterTrigger(input, config))
 				return false;
 
 			// we are configuring, the button is unpressed or the axis is relaxed
@@ -175,6 +177,9 @@ GuiInputConfig::GuiInputConfig(Window* window, InputConfig* target, bool reconfi
 				mHeldTime = 0;
 				mHeldInputId = i;
 
+				mAllInputs.clear();
+				mAllInputs.push_back(input);
+
 				return true;
 			}else{
 				// input up
@@ -184,9 +189,21 @@ GuiInputConfig::GuiInputConfig(Window* window, InputConfig* target, bool reconfi
 
 				mHoldingInput = false;
 
+				if (mHeldInput.type == InputType::TYPE_BUTTON)
+				{
+					auto altAxis = mAllInputs.where([&](auto x) { return x.device == mHeldInput.device && x.type == InputType::TYPE_AXIS; });
+					if (altAxis.size() >= 2)
+					{
+						auto groups = altAxis.groupBy([](auto x) { return x.id; });
+						if (groups.size() == 1)
+							mHeldInput = altAxis[0];
+					}
+				}
+
 				if(assign(mHeldInput, i))
 					rowDone(); // if successful, move cursor/stop configuring - if not, we'll just try again
 
+				mAllInputs.clear();
 				return true;
 			}
 		};
@@ -376,52 +393,7 @@ void GuiInputConfig::clearAssignment(int inputId)
 	mTargetConfig->unmapInput(GUI_INPUT_CONFIG_LIST[inputId].name);
 }
 
-bool GuiInputConfig::filterTrigger(Input input, InputConfig* config, int inputId)
+bool GuiInputConfig::filterTrigger(Input input, InputConfig* config)
 {
-#if defined(__linux__)
-	// on Linux, some gamepads return both an analog axis and a digital button for the trigger;
-	// we want the analog axis only, so this function removes the button press event
-
-	if(config->getDeviceNbAxes() >= 6)
-	{
-		// digital triggers are unwanted
-		if(input.type == TYPE_BUTTON
-		&& (GUI_INPUT_CONFIG_LIST[inputId].name == "l2"
-		|| GUI_INPUT_CONFIG_LIST[inputId].name == "r2"))
-		{
-			if(mHoldingInput && mHeldTime < 1000)
-			{
-				return false;
-			}
-			else
-			{
-				return true;
-			}
-		}
-	}
-
-	// ignore negative pole for axes 2/5 only when triggers are being configured
-	if(input.type == TYPE_AXIS)
-	{
-		if(GUI_INPUT_CONFIG_LIST[inputId].name == "l2"
-		|| GUI_INPUT_CONFIG_LIST[inputId].name == "r2")
-		{
-			if(input.value == 1)
-				mSkipAxis = true;
-			else if(input.value == -1)
-				return true;
-		}
-		else if(mSkipAxis)
-		{
-			mSkipAxis = false;
-			return true;
-		}
-	}
-#else
-	(void)input;
-	(void)config;
-	(void)inputId;
-#endif
-
 	return false;
 }
--- a/es-core/src/guis/GuiInputConfig.h
+++ b/es-core/src/guis/GuiInputConfig.h
@@ -6,6 +6,7 @@
 #include "components/ComponentGrid.h"
 #include "components/NinePatchComponent.h"
 #include "GuiComponent.h"
+#include "utils/VectorEx.h"
 
 class ComponentList;
 class TextComponent;
@@ -36,7 +37,7 @@ private:
 
 	bool assign(Input input, int inputId);
 	void clearAssignment(int inputId);
-	bool filterTrigger(Input input, InputConfig* config, int inputId);
+	bool filterTrigger(Input input, InputConfig* config);
 
 	void rowDone();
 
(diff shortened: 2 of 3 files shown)


dmanlfc:
We surely don't need these methods now...

InputManager::getAxisCountByDevice

InputManager::getButtonCountByDevice

InputManager::getDeviceGUIDString

dmanlfc:
Also, we shouldn't limit the id's if axis.

I would remove (input.id == 2 || input.id == 5)

That way controllers that may use different id numbers per trigger will also benefit.

TVsIan:
I'll make a test with those changes, the axis IDs were specifically to prevent negative inputs, but I'm not sure how common that would be. I should have that done some time tomorrow evening after work.

TVsIan:
Looks Ok as it is now, just tested and was able to map everything properly.

dmanlfc:
@nadenislamarre this works well. Can we merge & monitor at the start of the v35 cycle?

nadenislamarre:
it has to be validated by @fabricecaruso

nadenislamarre:
i'm really not fan of checking the number of axis and not the pads name. we will have false positive, for sure.
for me we should hardcode blacklist pads or do a more complex way.

fabricecaruso:
I don't really understand the mSkipAxis thing.
Plus : mSkipAxis is never initialized.

TVsIan:
That might be leftover RetroPie code, I’ll take a look at it tonight.

fabricecaruso:
You're right, you took the Retropie code... And the field is not initialized in their code neither...

TVsIan:
Yeah, a lot of it changed from the original version that just ported the
RetroPie code over though, theirs was only for the DS3 and one clone, we’ve
tried to make it more generic.

I can check and see if it’s initialized in another function, but a lot of
that can likely just get removed.

I think our timeout on holding a button to skip might be shorter than
theirs too, so I might need to adjust the fallback time as well. I haven’t
run into a controller that actually uses it though, I’ll have to try some
of my weird controllers.

dmanlfc:
> i'm really not fan of checking the number of axis and not the pads name. we will have false positive, for sure. for me we should hardcode blacklist pads or do a more complex way.

We do two checks, & a fallback so a false positive will be extremely rare. I have tested across 8 controllers, no issues. In addition managing controller names will require more maintenance for us.

If that rather rare occasion does ever eventuate, then you would use a controller name.

fabricecaruso:
I don't have such controllers to test...

Anyway : 
- It's not mergeable as long as the mSkipAxis is not initialized in the constructor.
- Retropie's code test the button ids  ( `if(input.type == TYPE_BUTTON && (input.id == 6 || input.id == 7))` )
The way they do seems preferable to what you do by testing the destination input config value ( using `GUI_INPUT_CONFIG_LIST[inputId].name == "l2"` ) :
If you want to set button 6 or 7 to l1 or r1 you'll meet the same exact problem.
- Also Retropie limits this for only a specific device (using Guid & Name). Maybe we should do that too. Not sure all the devices act this way...

TVsIan:
I can check the mSkipAxis and see what needs to be done.

I changed the check from button 6/7 to l2/r2 because buttons 6 and 7 may be
different buttons on other controllers, since we’re not filtering by guid
or name that could cause issues. Since the triggers would have analog
preferred, I changed it there.

I can change it back to controller IDs, but that will likely require
regular maintenance as new controllers with that issue are found.

fabricecaruso:
`I changed the check from button 6/7 to l2/r2 because buttons 6 and 7 may be different buttons`
Maybe, but the problem is not related to the destination entry in the config file but to the source button !

dmanlfc:
> `I changed the check from button 6/7 to l2/r2 because buttons 6 and 7 may be different buttons` Maybe, but the problem is not related to the destination entry in the config file but to the source button !

@fabricecaruso i think you misunderstand - he's only applying the axis preference when configuring l2 or r2.
the code checks you have 6 or more axis, if you do it's very likely the 5th & 6th axis is for the triggers.
if they are, we will use axis instead of button, if they're not then we fallback to the button id like we do today. this is very safe & logical.

fabricecaruso:
I don't misunderstand : This code is filtering the destination (the config file), not the source (the controller)....
Retropie's code filters the source. This code filters the destination.
In my opinion, it's not a good way to go : What if you want to configure the r2 button into r1 !? The problem occurs...

dmanlfc:
> I don't misunderstand : This code is filtering the destination (the config file), not the source (the controller).... Retropie's code filters the source. This code filters the destination. In my opinion, it's not a good way to go : What if you want to configure the r2 button into r1 !? The problem occurs... Is it a good solution ? not sure...

it doesn't filter the destination, it writes an alternative based on the trigger designation which on some controllers may be a button or an axis. it get's the button or axis from the controller...
why on earth would you want to configure r2 into r1? and even if you did, then new r1 would be a button (like the bumpers) and then the subsequent emulator controller config would leverage it.

here is the problem with retropie's method:

1. You have to maintain names & guids (there will be a lot more than what retropie has)
2. Some Sony controller candidates have names such as 'Wireless Controller' which leads to false identification as another controller can easily have that same name. Therefore potentially mapping the controller incorrectly altogether.
3. They assume the button id's for input & output - also not good for a wider range of controllers. Restricting the scope of the fix.

This patch has been tested vigorously in #beta with multiple users without maintenance. As outlined earlier, this should be applied at the beginning of the v35 release cycle & tested.

Please confirm we can merge...

fabricecaruso:
"it doesn't filter the destination" -> Yes it does... 
Anyway : Not as long as the mSkipAxis is not initialized.

nadenislamarre:
@dmanlfc
1=> You have to maintain names & guids => this is supposed to be a low number of devices. consider it as a bug. it cannot be automatic technically.
2=> so, use a list of vid/pid/name
3 => we must restrict the issue to the real issue. this fix is dangerous. it will come back in 2 months because we will have false positive, and we have nothing else to do that remove it completly to fix in a hurry.

fabricecaruso:
I may have a better idea than the "filterTrigger". If I understand well the problem. For this button, you receive TYPE_BUTTON, then TYPE_AXIS. As TYPE_BUTTON was received first, it's selected, but not the axis.

This is what I imagined which would be far more generic :
- When pressing the button & when setting mHoldingInput to true, we start recording all coming inputs in a vector ( mAllInputs ).
- When releasing the initial pressed button, we watch if it's a button ( InputType::TYPE_BUTTON ) then, we read all our recorded inputs.
- If in the recorded inputs we have at least 2 TYPE_AXIS from the device, and all the input are on the same id, then we can take the first TYPE_AXIS instead.

Can you tell me your point of view ( and/or test with affected controllers ? )

See screenshots of the code I wrote.

image

image

TVsIan:
> I may have a better idea than the "filterTrigger" thing.
> 
> If I understand well the problem. For this button with some controllers, you receive TYPE_BUTTON, then TYPE_AXIS. As TYPE_BUTTON was received first, it's selected at end, but not the axis. And You want to axis instead.
> 
> This is what I imagined which would be far more generic :
> 
> * When pressing the button & when setting mHoldingInput to true, we start recording all coming inputs in a vector ( mAllInputs ).
> * When releasing the initial pressed button, we watch if it's a button ( InputType::TYPE_BUTTON ) then, we read all our recorded inputs.
> * If, in the recorded inputs, we have at least 2 TYPE_AXIS from the device, and all the inputs are on the same id, then we can take the first TYPE_AXIS instead of the button.
> 
> Can you tell me your point of view ( and/or test with affected controllers ? )
> 
> See screenshots of the code I wrote.
> 

Reverting my changes and trying these instead gives me errors during compile:

```
In file included from src/guis/GuiInputConfig.h:9,

                 from src/guis/GuiDetectDevice.cpp:4:

src/utils/VectorEx.h: In function ‘std::vector<std::pair<decltype (fieldChooser(StructType())), std::vector<StructType> > > VectorHelper::groupBy(const std::vector<StructType>&, const FieldSelectorUnaryFn&)’:

src/utils/VectorEx.h:21:40: error: ‘find_if’ is not a member of ‘std’; did you mean ‘find’?

   21 |                         auto it = std::find_if(instancesByField.begin(), instancesByField.end(), [val](std::pair<FieldType, std::vector<StructType>>& x) { return x.first == val; });

      |                                        ^~~~~~~

      |                                        find


```

Followed by a lot of other failures that all seem to be related to using VectorEx, looks like it breaks other uses of the inputs.

fabricecaruso:
just add #include <algorithm> to VectorEx.h

TVsIan:
Testing this code with various controllers:

With analog triggers:
- DualShock 4 (Wired & Wireless): Axis
- XBox One S Controller (Wireless): Axis
- 8BitDo SN30 Pro+ (DualShock mode, wired): Axis, BUT on the first try, L1 came up as a button instead. When testing re-mapping the trigger, it alternated between button, axis, and -NOT DEFINED-. Pressing lightly seemed to trigger the button at first, but after 3-4 presses it was only not defined. Finishing and re-mapping the controller a second time, pushing the triggers all the way in mapped normally.

With digital triggers:
- 8BitDo SN30 Pro (DualShock mode, wired): Buttons
- PDP Versus Fighting Pad for PS3: Buttons
- Razer Panthera PS4 Fightstick: Buttons
- Hori Pokken Controller: Buttons
- Switch Pro Controller (Wireless): Buttons

So everything seems to work as normal, with some possible hiccups if the trigger is only partially pressed. It could also be the controller itself.

I'm happy closing my PR in favor of this code instead.

dmanlfc:
Works as expected with my additional controllers too. Logitech, Xbox 360 pad, 8-bit (Xinput) DS3. I think we have a solution to put out into the wild... Thanks @fabricecaruso

fabricecaruso:
All right, so how do we proceed ? You update your PR with the new code ?

TVsIan:
PR is updated to revert my changes and replace them with yours. I'll do a separate test for the logging this weekend and see what comes up.

fabricecaruso:
Ok then I'll wait for your test results

TVsIan:
Having some trouble building it:
```[ 40%] Building CXX object es-core/CMakeFiles/es-core.dir/src/LocaleES.cpp.o
/home/ian/Repos/batocera-emulationstation/es-core/src/LocaleES.cpp:18:45: error: expected unqualified-id before ‘{’ token
   18 | char* pgettext(char* context, char* msgid); {
      |                                             ^
make[2]: *** [es-core/CMakeFiles/es-core.dir/build.make:230: es-core/CMakeFiles/es-core.dir/src/LocaleES.cpp.o] Error 1
```
I had been building it via building Batocera and pointing the makefile at my repo, but I had to do a clean build (that's still in progress) so I was trying to build ES by itself and just copy it over when it was done.

fabricecaruso:
"libintl" libary ( and HAVE_INTL directive ) is probably missing.

TVsIan:
On the first try with the 8Bitdo Pro+ in Mac (DualShock) mode, pressing l2 or r2 part way skipped the mapping. Pressing l2 all the way mapped to an axis properly, but r2 would not map at all. Changing the controller to XInput mode worked fine. On every future try, the triggers wouldn't map at all.

Oddly, I had to do it twice, es.customargs=--debug in batocera.conf did not start it in debug mode, there was nothing in the logs and when I ssh'd in and checked top, it wasn't in the task info. When I got it running in debug mode, there didn't seem to be anything useful in the log. It's really strange, it's like it's not seeing the axis at all, but it's still skipping it as if I'm holding something that's already mapped. I don't get "already in use" if I go back and try to map it again though, it just says not assigned.

After a controller firmware update, I got better results - L2 mapped to button 6, but R2 mapped to Axis 5+. I remapped L2 a few times, it alternated between Button 6 and Not Defined before finally hitting Axis 2+.

I suspect this is just 8BitDo weirdness since a firmware update fixed it (sort of), but maybe the filter could use a little tweaking. It looks like if it sees Axis+ and Axis- it maps, but this controller is only sending Axis+ for the most part unless it releases during the mapping.

Here's the relevant part of the log (which is missing the ones that went to not defined):

```
2022-05-29 19:18:54	DEBUG	Processing inputs for l2 :
2022-05-29 19:18:54	DEBUG	 -2: Button 6= 
2022-05-29 19:18:54	DEBUG	 -2: Axis 2+= 
2022-05-29 19:18:54	DEBUG	 -2: Button 6 = 
2022-05-29 19:18:54	DEBUG	 -> Assigned input : 2: Button 6= 
2022-05-29 19:18:54	INFO	  Mapping [Button 6] -> l2
2022-05-29 19:18:55	DEBUG	Processing inputs for r2 :
2022-05-29 19:18:55	DEBUG	 -2: Button 7= 
2022-05-29 19:18:55	DEBUG	 -2: Axis 5+= 
2022-05-29 19:18:55	DEBUG	 -2: Axis 5- = 
2022-05-29 19:18:55	DEBUG	 -2: Button 7 = 
2022-05-29 19:18:55	DEBUG	 -> Assigned input : 2: Axis 5+= 
2022-05-29 19:18:55	INFO	  Mapping [Axis 5+] -> r2
2022-05-29 19:19:01	DEBUG	Processing inputs for l2 :
2022-05-29 19:19:01	DEBUG	 -2: Button 6= 
2022-05-29 19:19:01	DEBUG	 -2: Axis 2+= 
2022-05-29 19:19:01	DEBUG	 -2: Button 6 = 
2022-05-29 19:19:01	DEBUG	 -> Assigned input : 2: Button 6= 
2022-05-29 19:19:01	INFO	  Mapping [Button 6] -> l2
2022-05-29 19:19:04	DEBUG	Processing inputs for l2 :
2022-05-29 19:19:04	DEBUG	 -2: Button 6= 
2022-05-29 19:19:04	DEBUG	 -2: Axis 2+= 
2022-05-29 19:19:04	DEBUG	 -2: Axis 2- = 
2022-05-29 19:19:04	DEBUG	 -2: Button 6 = 
2022-05-29 19:19:04	DEBUG	 -> Assigned input : 2: Axis 2+= 
2022-05-29 19:19:04	INFO	  Mapping [Axis 2+] -> l2
```

dmanlfc:
@TVsIan delete your .sdl2 folder should be under /userdata/system then reboot & retry.

That usually fixes button mapping issues when axis is not registering.

TVsIan:
> @TVsIan delete your .sdl2 folder should be under /userdata/system then reboot & retry.
> 
> That usually fixes button mapping issues when axis is not registering.

That was it, worked perfectly on the first try. Looks like the change is good to go.

dmanlfc:
Good to go, let's get it in people's hands to verify with a wider array of controllers.

fabricecaruso:
You log is really strange, in some cas, there's only one Axis 2+ in the received inputs. There's no event to tell it's released (  Axis 2- ).
As we're waiting to 2 inputs ( pressed/released -> altAxis.size() >= 2 ), it takes the button.

PR #1211, opened 2022-06-05
bbh and bbhcotw are gun games

jdeath:
bbh and bbhcotw are light gun games


jdeath:
Ok. I think I fixed the file with the new one. sorry, not good with git. I'll make a new pr

PR #1217, opened 2022-06-08
New screensaver mode for arcade cabinets (to be reviewed)

FRANCKYNO:
Creation of a new screensaver mode dedicated to arcade cabinets.

The idea is to cycle through the games in the EmulationStation interface randomly like in HyperSpin by moving and spinning quickly and randomly through the list or the wheel of games and systems and stopping on a game to view marquee, image, video, information available then switch to another game.

I am not a computer scientist. I did my best to write the code but I need help to write the code cleanly and make it work properly. @fabricecaruso: I hope you can provide your help and expertise to set up this new screensaver. Thanks a lot

### DIFF
--- a/es-app/src/SystemScreenSaver.cpp
+++ b/es-app/src/SystemScreenSaver.cpp
@@ -182,22 +182,15 @@ void SystemScreenSaver::startScreenSaver()
 		mVideoChangeTime = Settings::getInstance()->getInt("ScreenSaverSwapBrowsingTimeout");
 		isBrowsingScreensaver = true; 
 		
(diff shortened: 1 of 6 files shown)


fabricecaruso:
Hi. It does not compile

fabricecaruso:
I think the it's almost impossible to write what you want to do in the current state. It would require very big changes & the code you initiated is far from accomplishing it. Methods like getFileDataEntries don't belong to the screensaver class... 
It would require to manage this with a view context & resolve problems when menus are open... 
It's would be a pretty big deal.

I'm closing this PR.

PR #1249, opened 2022-07-11
Introduce themable click actions + if condition for themes

fabricecaruso:
### Add themable click actions : 

Applies on text & image components : `<image onclick="back" [...]`
Note : Please bump the theme, there are 2 working & useful samples in it.

- Gamelist View onclick actions : 
```
	back
	options
	gameoptions
	launch 
	search
	savestates
	favorite
	random
```

- System View onclick actions : 
```
	netplay
	navigationbar
	search
	open
	random
	prev
	next
```
- use "none" to disable default click action for an element.

### Add if condition for theme elements, with a simple evaluator :

Applies to any element : ` <pos if="{system.name} == 'collections'" [...]` 

+ extend themes variable list. Variables are now :

		global.architecture
		global.help					( bool )
		global.clock					( bool )
		global.cheevos				( bool )
		global.cheevos.username
		global.netplay				( bool )
		global.netplay.username
		system.cheevos				( bool )
		system.netplay				( bool )
		system.savestates				( bool )
		system.fullName
		system.group
		system.hardwareType
		system.manufacturer
		system.name
		system.releaseYear
		system.releaseYearOrNull
		system.sortedBy
		system.theme
		system.command

### DIFF
--- a/es-app/src/ApiSystem.cpp
+++ b/es-app/src/ApiSystem.cpp
@@ -1250,6 +1250,8 @@ bool ApiSystem::isScriptingSupported(ScriptId script)
 	case ApiSystem::UPGRADE:
 		executables.push_back("batocera-upgrade");
 		break;
+	case ApiSystem::SUSPEND:
+		return Utils::FileSystem::exists("/usr/sbin/pm-suspend");
 	}
 
 	if (executables.size() == 0)
@@ -1714,5 +1716,11 @@ bool ApiSystem::emuKill()
 	LOG(LogDebug) << "ApiSystem::emuKill";
 
 	return executeScript("batocera-es-swissknife --emukill");
+}
+
+void ApiSystem::suspend()
+{
+	LOG(LogDebug) << "ApiSystem::suspend";
+	executeScript("/usr/sbin/pm-suspend");
 }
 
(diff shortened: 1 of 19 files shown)


pajarorrojo:
Thanks for your work, really. I've tried the new "if" method with variables and it works great, it's brilliant, congratulations! I think it will give us many new options for the development of layouts in the themes :). Just a suggestion, we can add to the list of variables some new ones like?:

global.cheevos.points
global.cheevos.rank
global.cheevos.friends

And to detect the screen resolution in ES, something like?.

global.aspect.ratio

Please 

I'm very excited about the new environment that opens with this new "if" method. Thanks again!

fabricecaruso:
Interesting for global.aspect.ratio. I'll think about it
Complex to write for global.cheevos.points/rank/friends, because it would require a connection before system & themes are loaded.

pajarorrojo:
The aspect ratio sounds good. About the cheevos, do you mean the actual username variable is not pinging the retroachievements server? Is it "just" the string that the user fills in the field on the configuration screen?

pajarorrojo:
Understood. Complex thing then... Thanks Fabrice!

pajarorrojo:
Would it be possible to extend the system onclick action with retroachievements? this click would open the retroachievements window that is now accessed from the main menu

fabricecaruso:
it already exists. Action is named cheevos

pajarorrojo:
great! thanks. Please, is there a site or file where we can see the complete list of actions?

PR #1268, opened 2022-07-25
When mp3 have both tags, use ID3v2 rather than ID3v1.

lbrpdx:
I can't recall if there's any reason why we parsed ID3 tags v1 before v2 when both are available in an mp3. If anyone has an idea, please let me know.
Otherwise, more recent tags are probably a better choice (first, because 255 characters maximum are better than 30)

### DIFF
--- a/es-core/src/AudioManager.cpp
+++ b/es-core/src/AudioManager.cpp
@@ -414,7 +414,31 @@ void AudioManager::playSong(const std::string& song)
 
 	LOG(LogDebug) << "AudioManager::setSongName";
 
-	// First, start with an ID3 v1 tag	
+	// First let's try with an ID3 v2 tag
+#define MAX_STR_SIZE 255 // Empiric max size of a MP3 title
+
+	ID3v2_tag* tag = load_tag(song.c_str());
+	if (tag != NULL)
+	{
+		ID3v2_frame* title_frame = tag_get_title(tag);
+		if (title_frame != NULL)
+		{
+			ID3v2_frame_text_content* title_content = parse_text_frame_content(title_frame);
+			if (title_content != NULL)
+			{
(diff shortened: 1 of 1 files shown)


fabricecaruso:
I can't remember why, but probably because is was sometimes crashing.

Anyway. There's no garbage collector in C++ and there are memory leaks :
- Every call to **load_tag** leads to memory leaks -> We should add calls to free_tag(tag);
- Every call to **parse_text_frame_content** leads to memory leaks too -> There's no free method for it, so we'd have to write one & call it.

As long as these leaks occur, I'm not sure, using Idv2 is a good thing.

lbrpdx:
Thanks for your review @fabricecaruso. This new push should fix the mem leak + added "artist" parsing and addition to the string displayed if there's one.

fabricecaruso:
- You don't "setSongName(song_name) / return" if there is no artist tag.
- if (title_content->size >0) -> even if it's not probable, you should continue to test NULL to secure the code -> ` if (title_content != NULL && title_content->size >0)`
- title_content->data is not freed & is still leaking. The same with artist_content->data. You should `free(title_content->data);` too

lbrpdx:
Github lost my previous comments twice... so putting here it again.
- For `title_content->data` I (wrongly) thought it was taken care of by freeing `title_content` but I forgot how old-school C++ is with regards to memory management :(
- `if (title_content != NULL)` : I added it but I think it's useless as the only case it can return NULL is when `frame == NULL` in the code for `parse_text_frame_content`, which was tested beforehand:
 https://github.com/batocera-linux/batocera-emulationstation/blob/1dd4f5f86050d412901e29cb7c42e01aa1190c3d/external/id3v2lib/src/frame.c#L71
 - Finally, I also checked that the `ID3v2_frame` were freed when `free_tag()` is invoked.

PR #1275, opened 2022-08-07
more accurate descriptive label for video mode

Hew-ux:
Complement to https://github.com/batocera-linux/batocera.linux/pull/6811

"The video mode description says 'Force the emulator to run at this resolution', but that's not entirely true. The emulator's rendering resolution is what decides your game's resolution, the video mode is just the resolution of the signal sent to the display.

The reason for this change was a new user thinking that video mode was the option to change their rendering resolution. This change aims to improve the new user experience for this setting."


fabricecaruso:
"Send this resolution to the display". You really think it's more accurate ?
"Sets the screen resolution" simply.... No ?

Hew-ux:
That's better actually, let's use that instead.

fabricecaruso:
"to this" ??? Why "to this" ?

Hew-ux:
Because you're setting the display resolution to this, notably in comparison to "that", the system's native resolution. But you're right, they aren't strictly necessary words, so let's remove them.

PR #1278, opened 2022-08-12
update gun games

nadenislamarre:


Darknior:
Thanks for the update.
Just consider on NES you have listed two times "russianroulette"

PR #1283, opened 2022-08-18
rename to new gun settings menus

Hew-ux:
a bit more descriptive as to what these new menus are for

### DIFF
--- a/es-app/src/guis/GuiMenu.cpp
+++ b/es-app/src/guis/GuiMenu.cpp
@@ -2391,7 +2391,7 @@ void GuiMenu::openControllersSettings(int autoSel)
 	  if (gun->needBorders())
 	    sindenguns_menu = true;
 	if(sindenguns_menu) {
-	  s->addEntry(_("SINDEN GUNS"), true, [this] { openControllersSpecificSettings_sindengun(); });
+	  s->addEntry(_("SINDEN GUN SETTINGS"), true, [this] { openControllersSpecificSettings_sindengun(); });
 	}
 #endif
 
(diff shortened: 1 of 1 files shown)


nadenislamarre:
"WIIMOTE AS LIGHT GUN SETTINGS" => in case the wiimote is as the joystick, this menu is still visible to swich it as a gun and vice and versa. this is not wiimote as light gun settings.
"WHITE BORDER SIZE" => it could be change. it is really gun border size.

Hew-ux:
Still arguably better than "WIIGUN", which does not have meaning nor indication of what the menu is used for. If you have a better suggestion I'm open to changing it.

The second point is fair enough, but the reason I would suggest changing it is that while border is immediately understandable what it is referring to. The proper label would be "SINDEN GUN TRACKING BORDER", but that would be to long. "GUN BORDER" does not make sense as gun is not a proper noun.

nadenislamarre:
wiigun is the name i gave to the batocera gun wiimote system.
this is the name that appear when you plug a wiimote as a gun : "Wiigun plugged" on top on es.
It was how i named it in the wiki too before it was renamed 2 days ago too.
If the name is removed everywhere it is, sure, nobody will understand this word.
it  may be called the menu wiimote but this is the section for guns, so, let's call a cat : a cat.

"gun borders" makes completly sense.
on the windows software, the name is named just "borders"
This string is visible once you're already in the menu "SINDEN GUN" menu, so, repeating "SINDEN GUN" as prefix of each option label is not required. You can rename it "borders" if you prefer.

MENU "SINDEN GUNS"
     OPTION "GUN BORDERS"
     OPTION "RECOIL"
 MENU "WIIGUNS"
     OPTION "MODE"

Hew-ux:
When introducing a new concept, especially one that's not been used previously, it needs to be adamantly clear that is what is happening. This was not the case, as evidenced by me not thinking that was a new terminology. I'll rectify this in the wiki.

With that said, I still think the term "Wiiguns" is ambiguous. "Wiimote gun" seems like it would cause less confusion.

About the Sinden, there are people buying these light guns to use in Batocera. They wouldn't know about what the control panel in Windows would say about that setting.

But you are right, there is no need for the redundant labelling. I'll make the adjustments next time I'm able to.

nadenislamarre:
ok for wiimote gun. or even wiimote while it helps to switch to joystick.

PR #1287, opened 2022-08-25
Workaround for mouse cursor shown on start

NostalgicDavid:
Ignore initial SDL_MOUSEMOTION events with zero coordinates

Fixes https://github.com/batocera-linux/batocera.linux/issues/6920

### DIFF
--- a/es-core/src/Window.cpp
+++ b/es-core/src/Window.cpp
@@ -35,7 +35,7 @@ Window::Window() : mNormalizeNextUpdate(false), mFrameTimeElapsed(0), mFrameCoun
 	mBackgroundOverlay->setImage(":/scroll_gradient.png"); 
 
 	mSplash = nullptr;
-	mLastShowCursor = -1;	
+	mLastShowCursor = -2;
 }
 
 Window::~Window()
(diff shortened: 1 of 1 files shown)


NostalgicDavid:
@fabricecaruso Can you please review?

nadenislamarre:
@NostalgicDavid can you explain the initial problem ?

NostalgicDavid:
@nadenislamarre When Batocera v35 beta boots in Odroid GO Advance mouse cursor is shown in Emulation station for 5 seconds.
Odroid GO Advance is handheld without mouse without touchscreen. Analog joystick don't move mouse cursor. Mouse cursor was not shown in Batocera v34.
Not tested in other devices.

fabricecaruso:
With the proposed fix mLastShowCursor is always equal to -2 & its value is definitive, so, the cursor will never show anymore.
So : it's a bad fix

NostalgicDavid:
@fabricecaruso **I have tested! The cursor is shown when the mouse is moved.**

1. When cursor coordinates is not (x=0, y=0) then cursor is shown and 0 is stored in mLastShowCursor. 
1. When cursor is not moved for 5 seconds -1 is stored in mLastShowCursor ([line 413 is not changed](https://github.com/batocera-linux/batocera-emulationstation/blob/406352dbbaf51b8aadd75de75c1bfc834daadd6a/es-core/src/Window.cpp#L413)).
1. If cursor is moved back to (x=0, y=0) then cursor is still shown, because mLastShowCursor is not -2.

Can you please review more careful?

fabricecaruso:
1. mmm yes it's a || condition, so it can change when x/y != 0
2. nope, when it's -2 it's skipped because of if (mLastShowCursor >= 0), but it's not a problem.
3. yep

Indeed it should work because it's a or condition. I don't know why I read it like if it was &&

NostalgicDavid:
Thank you!

PR #1292, opened 2022-08-29
mapper input labels to be homogenuous

Hew-ux:
Less dependent on the user knowing the layout of both the SNES and PS3/4 controller.

Uses the same co-ordinate system for face buttons as used throughout the rest of Batocera:
screenshot-2022 08 26-21h45 38

Replaces L1/R1 with Left shoulder/Right shoulder and L2/R2 with Left trigger/Right trigger.

Replaces L3/R3 with Left stick press/Right stick press.

Specifies the "up/down/left/right" are for the D-pad.

### DIFF
--- a/es-app/src/guis/GuiKeyMappingEditor.cpp
+++ b/es-app/src/guis/GuiKeyMappingEditor.cpp
@@ -34,8 +34,8 @@ void GuiKeyMappingEditor::initMappingNames()
 		{ "start",   "START",        ":/help/button_start.svg" },
 		{ "select",  "SELECT",       ":/help/button_select.svg" },
 
-		{ "a",				 InputConfig::buttonLabel("a"),    InputConfig::buttonImage("a") },
-		{ "b",				 InputConfig::buttonLabel("b"),    InputConfig::buttonImage("b") },
+		{ "a",               "EAST",    ":/help/buttons_east.svg" },
+		{ "b",               "SOUTH",    ":/help/buttons_south.svg" },
 
 		{ "x",               "NORTH",   ":/help/buttons_north.svg" },
 		{ "y",               "WEST",    ":/help/buttons_west.svg" },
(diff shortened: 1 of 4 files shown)


fabricecaruso:
I can understand ( and agree ) with the North/Sound thing, but I personnaly hate the "[...] ON SNES" thing.

Hew-ux:
To be honest, I just kept that there in case you wanted to keep the original abxy. I would prefer to not have the "... ON SNES" too, so I'll just remove them?

Hew-ux:
Now it looks like this:
screenshot-2022 08 31-19h00 08
screenshot-2022 08 31-19h01 05
Tested and ready to merge.

Hew-ux:
Any reason this hasn't been merged yet?

Tartifless:
Hi, is it possible that this generated inconsistency between A & B buttons.
Now if we configure controllers according to the layout, EAST becomes the selection/validation button, and SOUTH brings you back from gamelist to system view.

Which is inconsistent with what is displayed at the bottom of the theme.

This is without A & B inversion set in developers options.

Hew-ux:
@Tartifless  That would be strange, as the only things I changed were the label strings, not any functionality.

Considering the time difference between "when this PR was made" and "when it was merged" was half a year, it is possible that ES now behaves differently. I will investigate, thank you.

Hew-ux:
I've just tested the latest build, and yes it works fine for me.

It's still possible that there was an issue, which fabrice has seemingly already made a fix for: https://github.com/batocera-linux/batocera-emulationstation/commit/a5dfcd0c85629b6906f2788c2404878dfe7576e4 so check that out when it gets built.

fabricecaruso:
works fine for you because you did'nt test with Windows. A/B Buttons display & are inverted with Windows build
I fixed this.

PR #1293, opened 2022-08-30
add calibration to menu

Hew-ux:
Adds recalibration function to the Wiimote gun menu.

Need to wait on confirmation that this will work before merging.

### DIFF
--- a/es-app/src/ApiSystem.cpp
+++ b/es-app/src/ApiSystem.cpp
@@ -1717,3 +1717,7 @@ void ApiSystem::replugControllers_wiimotes() {
   LOG(LogDebug) << "ApiSystem::replugControllers_wiimotes";
   executeScript("/usr/bin/virtual-wii-mouse-bar-remap");
 }
+
+bool ApiSystem::wiigun_calibrate() {
+	return executeScript("/usr/bin/batocera-gun-calibrator-daemon $env{DEVNAME} 260");
+}
--- a/es-app/src/guis/GuiMenu.cpp
+++ b/es-app/src/guis/GuiMenu.cpp
@@ -2568,6 +2568,12 @@ void GuiMenu::openControllersSpecificSettings_wiigun()
 	    ApiSystem::getInstance()->replugControllers_wiimotes();
 	  }
 	});
+
+	s->addEntry(_("RECALIBRATE WIIMOTE GUN"), false, [this, s]
+	{
+		wiigun_calibrate();
+	});
+
 	mWindow->pushGui(s);
 }
 


nadenislamarre:
this is not correct. a end user must not call this script

Hew-ux:
@nadenislamarre Could you please explain? Are you suggesting that the user is not meant to calibrate their controller?

I've copied verbatim the script call used when the user presses the HOME button on the wiimote for three seconds. This is another way to access that functionality; for better accessibility (so that you aren't dependent on reading the wiki in order to know this).

nadenislamarre:
no, this is not what is called when pressed 3 sec. no script is called when pressing 3s.
this script is the driver over the wiimote driver to add the calibration.
it is started by udev when plugin the wiimote.
DEVNAME is an env variable set by udev. here, it refers to nothing.

Hew-ux:
Okay, so what is the command which is run when the user holds the button for 3s? Are you saying it is impossible to have a button in the menu perform this same action?

nadenislamarre:
all is possible. it is just software.
no command is run when the button is pressed 3 seconds.
the driver is just run when the gun is plugged. and when you press 3s, the running driver detects it and move the mouse.

https://github.com/batocera-linux/batocera.linux/blob/master/package/batocera/controllers/batocera-gun-calibrator/batocera-gun-calibrator#L212

Hew-ux:
So you're saying it's just been coded into the script which runs at all times in the background, and that's what allows the controller to begin calibration? If so, how do we separate that function outside of that background service and have it be callable externally?

Could we make it a function within that program, which could then be called?

Darknior:
> So you're saying it's just been coded into the script which runs at all times in the background, and that's what allows the controller to begin calibration? If so, how do we separate that function outside of that background service and have it be callable externally?
> 
> Could we make it a function within that program, which could then be called?

It's why i have write you to only show a popup to inform the user, he will push the B or Home button alone and it will launch the script. We must only explain him how to do.
I think it is the best solution and there is nothing to code.

Darknior:
I'm agree with you, but when we configure a gamepad on batocera we can see a pop-up and it is really perfect.
And we must press a button the gamepad we want to configure.
I only propose to make same with guns. And for guns there is more to explain. 3 points very important for users.
For me it is not a band-aid solution. It should be the best intuitive solution.

Hew-ux:
I am against the notion of that pop-up. Binding a controller should be self-explanatory. That's what the purpose of the label change PR I've made here is for: https://github.com/batocera-linux/batocera-emulationstation/pull/1292

Darknior:
> I am against the notion of that pop-up. Binding a controller should be self-explanatory. That's what the purpose of the label change PR I've made here is for: #1292

Your PR is excellent for sure, i like it so much.
But it come after the pop-up. And why ? because we must explain the user that he can push a button on the controller he want to configure it. There is no other solution and it is the better way. It's easy to understand and to use.
No one will never read Wiki, the best way is to implement more important informations in ES.
Many cool commercial systems are using this method ... 
For sure having so many popup every where, or so many things to read is boring ... 
But for game pad it is the best way and same for Guns. Because you only think to calibrate the gun one time, but you don't think to explain the user the two other points he must know.
Play a lot with guns, try them on many games ... you will see it is important to know you can calibrate them every where every time, for exemple. 
I don't tell it to you to boring you, i love your work, i only try to help you on this feature. And for guns we have speak a lot with @Tovarichtch and it is the best way to reduce questions on Discord.

Tovarichtch:
I confirm that it's a bit common to not read the wiki and try to play right away without calibrating first. I believe this is because most people use the dolphinbar which doesn't have any calibration. Changing this behavior through Batocera v35 might require a popup or at the very least a notification in the middle just like it does when plugging in a gun/controller/etc. So when the gun is plugged, at least notify the user to calibrate. Like :

"IMPORTANT : hold Home button to calibrate"

Something simple and short.

PR #1295, opened 2022-09-06
Update emulationstation2.po

joao-pedro-barbosa:
File update and cleaning from master file .pot.
Other small fixes.


nadenislamarre:
hum conflict. i think you did not start from the last file that i fixed.
eventually, if you don't know how to fix, send me the file via discord

joao-pedro-barbosa:
Hi @nadenislamarre. I began from firt line and compared all to the end, with .pot file that I suposed is the guide. The existing file previous my change is outdated (The are obsolete strings comparing it with .pot, and other (like soft points) that are not present).
If you need I can check the new file with the last beta you release.

PR #1311, opened 2022-10-05
Support battery readings from standalone fuel gauge and charger

theophile:
The Linux kernel has drivers for standalone battery-monitoring hardware that does not show up in /sys/class/power_supply as a discrete battery device with "status" and "capacity" properties. For example, in my use case, I am using the LTC2941 fuel gauge and the generic "gpio-charger" driver to detect the presence of +5V on the USB charging port. The LTC2941 does not report "capacity" but it does report the current charge level and the max charge level of the battery. Meanwhile, the gpio-charger driver does report the charging status via a "status" property, but it's a separate device from the fuel gauge.

This PR adds support for such devices so that the battery status overlay works as intended. If I did it right, existing functionality is preserved for users without such hardware, or with a standard battery device. ES first checks for a battery device and if it finds one, it is read as before. If it does not find one, it instead looks for devices in /sys/class/power_supply/ that contain "fuel" or "charger" in the device name. If it finds them, it calculates the remaining battery capacity from the "charge_now" and "charge_max" properties, and it determines the charging state from the charger's "status" property.

### DIFF
--- a/es-core/src/platform.cpp
+++ b/es-core/src/platform.cpp
@@ -367,6 +367,12 @@ BatteryInformation queryBatteryInformation()
 				batteryCurrChargePath = fuelgaugeRootPath + "/charge_now";
 				batteryMaxChargePath = fuelgaugeRootPath + "/charge_full";
 				batteryCapacityPath = ".";
+				// If there's a fuel gauge without "charge_now" or "charge_full" property, don't poll it
+				if ((!Utils::FileSystem::exists(batteryCurrChargePath)) || (!Utils::FileSystem::exists(batteryMaxChargePath)))
+				{
+					batteryCurrChargePath = ".";
+					batteryMaxChargePath = ".";
+				}
 			}
 			if (!chargerRootPath.empty())
                                 batteryStatusPath = chargerRootPath + "/status";
@@ -378,10 +384,9 @@ BatteryInformation queryBatteryInformation()
 			batteryStatusPath = batteryRootPath + "/status";
 			batteryCapacityPath = batteryRootPath + "/capacity";
 		}
-
 	}
 
-	if (batteryStatusPath.length() <= 1)
+	if ((batteryStatusPath.length() <= 1) || (batteryCurrChargePath.length() <= 1))
 	{
 		ret.hasBattery = false;
 		ret.isCharging = false;
@@ -401,7 +406,7 @@ BatteryInformation queryBatteryInformation()
 			ret.level = int(round((now/full)*100));
 		}
 		else
-			ret.level = atoi(Utils::FileSystem::readAllText(batteryCapacityPath).c_str());
+			ret.level = Utils::String::toInteger(Utils::FileSystem::readAllText(batteryCapacityPath).c_str());
 	}
 
 	return ret;
--- a/es-core/src/platform.cpp
+++ b/es-core/src/platform.cpp
@@ -336,29 +336,49 @@ BatteryInformation queryBatteryInformation()
 
 	static std::string batteryStatusPath;
 	static std::string batteryCapacityPath;
+	static std::string batteryCurrChargePath;
+	static std::string batteryMaxChargePath;
 
 	// Find battery path - only at the first call
 	if (batteryStatusPath.empty())
 	{
 		std::string batteryRootPath;
+                std::string fuelgaugeRootPath;
+                std::string chargerRootPath;
 
 		auto files = Utils::FileSystem::getDirContent("/sys/class/power_supply");
 		for (auto file : files)
 		{
-			if (Utils::String::toLower(file).find("/bat") != std::string::npos)
-			{
+			if ( (Utils::String::toLower(file).find("/bat") != std::string::npos) && (batteryRootPath.empty()) )
 				batteryRootPath = file;
-				break;
-			}
-		}
 
+			if ( (Utils::String::toLower(file).find("fuel") != std::string::npos) && (fuelgaugeRootPath.empty()) )
+				fuelgaugeRootPath = file;
+
+                        if ( (Utils::String::toLower(file).find("charger") != std::string::npos) && (chargerRootPath.empty()) )
+                                chargerRootPath = file;
+                }
+
+		// If there's no battery device, look for discrete charger and fuel gauge
 		if (batteryRootPath.empty())
-			batteryStatusPath = ".";
+		{
+			if (!fuelgaugeRootPath.empty())
+			{
+				batteryCurrChargePath = fuelgaugeRootPath + "/charge_now";
+				batteryMaxChargePath = fuelgaugeRootPath + "/charge_full";
+				batteryCapacityPath = ".";
+			}
+			if (!chargerRootPath.empty())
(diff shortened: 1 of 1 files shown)


es-core/src/platform.cpp
```diff
@@ -370,8 +390,18 @@ BatteryInformation queryBatteryInformation()
...
-		ret.level = atoi(Utils::FileSystem::readAllText(batteryCapacityPath).c_str());
+		std::string chargerStatus;
+		chargerStatus = Utils::String::replace(Utils::FileSystem::readAllText(batteryStatusPath), "\n", "");
+		ret.isCharging = ((chargerStatus != "Not charging") && (chargerStatus != "Discharging"));
+		// If reading from fuel gauge, we have to calculate remaining charge
+		if (batteryCapacityPath.length() <= 1)
+		{
+			float now = std::stof(Utils::FileSystem::readAllText(batteryCurrChargePath).c_str());
+			float full = std::stof(Utils::FileSystem::readAllText(batteryMaxChargePath).c_str());
+			ret.level = int(round((now/full)*100));
```
fabricecaruso, line 401: Potential division by 0 if batteryMaxChargePath does not exist or if the result is empty ( or not convertable to float )
Also Prefer Utils::String::toFloat instead of std::stof

es-core/src/platform.cpp
```diff
@@ -370,8 +390,18 @@ BatteryInformation queryBatteryInformation()
...
+		}
+		else
+			ret.level = atoi(Utils::FileSystem::readAllText(batteryCapacityPath).c_str());
```
fabricecaruso, line 404: Prefer Utils::String::toInteger()

es-core/src/platform.cpp
theophile: Good catch, thanks. I added a check to make sure that the "charge_now," "charge_full," and "status" files all exist. If any of them is missing, it should report no battery. This should also prevent the possibility of division by zero because if "charge_full" exists, it should never be zero.

theophile: Also, I tried to use Utils::String::toFloat instead of std::stof, but when I do that the battery overlay always reads 0%. I am not familiar enough with C++ to figure out why it works with std::stof but not with Utils::String::toFloat.

es-core/src/platform.cpp
theophile: The existing code uses atoi here, but I went ahead and changed it to Utils::String::toInteger. I should note that I do not have a device with a standard battery to test this on, so I haven't been able to verify whether this change somehow breaks the existing battery gauge.

theophile:
Added an additional commit to address comments from @fabricecaruso

es-core/src/platform.cpp
fabricecaruso: What's the decimal separator in the output you have for Utils::FileSystem::readAllText(batteryMaxChargePath) ?

theophile: > What's the decimal separator in the output you have for Utils::FileSystem::readAllText(batteryMaxChargePath) ?

The content of /sys/class/power_supply/fuelgauge/charge_full shouldn't ever have a decimal because it's an integer: https://github.com/torvalds/linux/blob/2c85ebc57b3e1817b6ce1a6b703928e113a90442/drivers/power/supply/ltc2941-battery-gauge.c#L90

As an example, my battery's charge_full is 5570475

fabricecaruso: Why stof if it's an integer ? just use integers, then :
int now = Utils::String::toInteger(Utils::FileSystem::readAllText(batteryCurrChargePath).c_str());
int full = Utils::String::toInteger(Utils::FileSystem::readAllText(batteryMaxChargePath).c_str());
ret.level = int(round((now*100)/full));

theophile: Because the floating point calculation is able to handle rounding better. My charge circuit doesn't always get the battery all the way up to 5570475, so even if it's at 5570470, with the simple integer math, the battery gauge only says 99%.

fabricecaruso: I don't see how you secured the potential divide by zero thing...

theophile: If /sys/class/power_supply/charge_max exists, it will not be non-zero. If it does not exist, this will report no battery and the math doesn't get done.

fabricecaruso: and what if Utils::FileSystem::readAllText fails ? and what if atoi fails ?

theophile: The existing code already uses Utils::FileSystem::readAllText and atoi here to read sysfs node properties. I'm not sure I understand what it would mean for these functions to fail. Perhaps the result would be a battery indicator that always says 0%, as was the case when I tried to use Utils::String::toFloat instead of std::stof. I do not know for certain.

fabricecaruso: ReadAllText can fail if the file does not exist or is exclusively locked. Then you have 0.
atoi can fail if there's a non-number char in the result. (letter, point). Then you have 0.
If you have 0 in "full" : you perform a division by 0, and then the program crashes...
It probably will never happen, but it's alway a good practice to secure that kind of things. ( if full != 0 ... )

miwasp:
I just tested this on an Odroid Go Super and it does not work. There is no battery status at all.
The reason is line 389 in platform.cpp:
`if ((batteryStatusPath.length() <= 1) || (batteryCurrChargePath.length() <= 1))`
The batteryStatusPath contains the correct path but batteryCurrChargePath is never filled with any value, so ES thinks the device has no battery. If i change this line back to
`if ((batteryStatusPath.length() <= 1))`
everything works as before.

theophile:
> I just tested this on an Odroid Go Super and it does not work. There is no battery status at all. The reason is line 389 in platform.cpp: `if ((batteryStatusPath.length() <= 1) || (batteryCurrChargePath.length() <= 1))` The batteryStatusPath contains the correct path but batteryCurrChargePath is never filled with any value, so ES thinks the device has no battery. If i change this line back to `if ((batteryStatusPath.length() <= 1))` everything works as before.

If you have a moment, please try again with that line changed to:

`if ((batteryStatusPath.length() <= 1) && (batteryCurrChargePath.length() <= 1))`

Here's what I was trying to do. The original code uses the batteryStatusPath variable and if no battery device is detected, it defines that variable as "." Then later if the length of that variable is 1 character or less, it reports no battery. I tried to borrow the same logic using a new batteryCurrChargePath variable. So that line should be checking to see if batteryStatusPath batteryCurrChargePath are _both_ 1 character or less, and only then reporting no battery. I see that by using the || operator I was only checking to see if _either_ variable was 1 character or less, which would result in reporting no battery unless there is also a separate fuel gauge device, which of course was not the intention.

miwasp:
`if ((batteryStatusPath.length() <= 1) && (batteryCurrChargePath.length() <= 1))`
brings the battery status back on the Odroid Go Super.
Sounds like this is the way it should be.

theophile:
Thanks for confirming. I'll open another PR this weekend to fix.

Issue #1102, opened 2022-01-06
Internet Status shows "not connected" even when connected + can't update because of this


Hew-ux:
In **MAIN MENU** -> **NETWORK SETTINGS**, the **INTERNET STATUS** always says "NOT CONNECTED", even if there is an IP address available and showing. The internet connection itself is fine, I'm able to download themes, scrape for media, etc. successfully.

I theorise that because of the internet status claiming to be not connected, that the menu tool for upgrading also fails. So, boot up Batocera, it has a notification that an update is available (I assume that it checks the internet for that) and attempting to download the update results in "NOT CONNECTED", failing the update. It is still possible to update using `batocera-upgrade` via the command line, so this issue is purely with ES's method of upgrading.

Here's the result of `ifconfig`:
![Capture](https://user-images.githubusercontent.com/67527064/148334959-aa299552-4415-4949-ba8d-77386a83e4d6.PNG)

This happens no matter if I'm using a wireless connection, an ethernet cable, both at once, only one at a time, on PC/RPi3/RPi4/s905x3. It's all the exact same issue.

fabricecaruso:
Probably not directly related to ES. This piece of code in ES is very old now. It's using getifaddrs. Something must have changed at system level. The only way to know is to modify code & place logs in queryIPAdress() located in es-core/platform.cpp.
Can you do such a thing or not ?

Hew-ux:
Just tried making my own build of ES to test such changes:
![screenshot-2022 03 21-17h04 32](https://user-images.githubusercontent.com/67527064/159212165-a08373c4-e171-456b-9c19-5177f441cbee.png)
Now here's the funny thing, I have not made any modifications. Was there a PR made to fix this which didn't tag this issue, or was this fixed merely by me compiling my own version of ES?

Hew-ux:
Yep, went back to the regular beta and it seems to be functional now. Therefore, this issue shall be closed!

Frontalmule:
What did you do in order to solve your problem. It seems like it solved it self. I am having what appears to be the same issue. I have tried on RPi4 and N2+. With older versions and using the Beta version and I cannot get Connected.

Hew-ux:
@Frontalmule  I'm sorry to say but it really was like that. Considering that ES uses an old method of detecting an internet connection, it seems to be really inconsistent. Maybe it will one day start working for you too.

At the least, you could use the command line to update.

Hew-ux:
[...]

PR #1320, opened 2022-10-10
update ping utility

Hew-ux:
Check for the HTTP URL of Google instead of only relying on its DNS. This will improve the robustness of the tool and also how quickly it resolves (in my personal testing at least, pinging the http domain is 2ms quicker).

Related issue: https://github.com/batocera-linux/batocera-emulationstation/issues/1102

Also made the structure of the code more consistent with the rest of the file.

### DIFF
--- a/es-app/src/ApiSystem.cpp
+++ b/es-app/src/ApiSystem.cpp
@@ -332,8 +332,16 @@ std::pair<std::string, int> ApiSystem::scrape(BusyComponent* ui)
 
 bool ApiSystem::ping() 
 {
-	if (!executeScript("timeout 1 ping -c 1 -t 1000 8.8.8.8")) // ping Google DNS
-		return executeScript("timeout 2 ping -c 1 -t 2000 8.8.4.4"); // ping Google secondary DNS & give 2 seconds
+	// ping Google, if it fails then move on, if succeeds exit loop and return "true"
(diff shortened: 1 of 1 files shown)


es-app/src/ApiSystem.cpp
```diff
@@ -332,8 +332,16 @@ std::pair<std::string, int> ApiSystem::scrape(BusyComponent* ui)
 
 bool ApiSystem::ping() 
 {
-	if (!executeScript("timeout 1 ping -c 1 -t 1000 8.8.8.8")) // ping Google DNS
-		return executeScript("timeout 2 ping -c 1 -t 2000 8.8.4.4"); // ping Google secondary DNS & give 2 seconds
+	// ping Google, if it fails then move on, if succeeds exit loop and return "true"
+	if (!executeScript("timeout 1 ping -c 1 -t 1000 google.com"))
+	{
+		// ping Google DNS
+		if (!executeScript("timeout 1 ping -c 1 -t 1000 8.8.4.4"))
```
fabricecaruso, line 339: Should be 8.8.8.8, no ? You try 2 times the same 8.8.4.4....

Hew-ux: You're right, I meant that to be 8.8.8.8 and 8.8.4.4 so we wouldn't lose any functionality, but guess i got too caught up in the structure and testing. I'll make a commit to fix that soon.

PR #1347, opened 2022-11-11
Update gungames.xml

Tovarichtch:
Adding PS3 known light gun games


Tovarichtch:
Not finished at all, but let's start with this.

PR #1349, opened 2022-11-18
eu_ES Basque translation

devilschile2:
Missing translation have been produced


nadenislamarre:
thanks.
note that options are in a separate files too if you want to translate 100%

devilschile2:
Thank you for noticing. 
I will try to find a moment to do it.

PR #1353, opened 2022-11-26
FileFilter add Chinese Pinyin initial letter search

jjnetcn:
In China, there is a popular fast text search method called "Chinese Pinyin initial letter search" ( https://en.wikipedia.org/wiki/Pinyin ) in tvbox remote control.
for example:
街头霸王(Street Fighter)  Pinyin is:  Jie Tou Ba Wang,   we can filter by  "jtbw", "jtb" or "bw"
重装机兵(Metal Max) Pinyin is: "Zhong Zhuang Ji Bing",  can filter by "zzjb"  or "jb"
BTW: Some Chinese characters have multiple Pinyin, "重"can speak "Zhong" or "Chong",
as special, we can also filter by "czjb"

Here use hash table+std::search(KMP) fast algorithm and check all char <0x80
there is slight performance loss for ANSI code page, but hurge improvement for chinese charset


fabricecaruso:
Could this be activated only when the main language is Chinese ?
For other people (European, Americans) they won't see the difference between containsIgnoreCasePinyin && containsIgnoreCase.

PR #1368, opened 2022-12-08
Add Manufacturer & Type Sort

TVsIan:
This adds an additional sort type to the Systems list, tentatively "Manufacturer and Type," internally "subgroup."

The sort method is different than Sort by Manufacturer in that within the manufacturer's systems they get sorted by type before year. So for example, Nintendo will list all their consoles from NES to Wii U, then handhelds from Game & Watch to 3DS. Sega will list their arcade systems, then consoles, then the Game Gear. Atari is consoles, then computers, then the Lynx.

It was requested on the Batocera Discord, and it does make logical sense as a sort method as it keeps similar systems grouped together. The one outlier in Batocera is that Windows is tagged as "pc" rather than "computer," so Microsoft will go MSX - XBox - Windows. If this gets merged, I'll re-tag it as computer.

### DIFF
--- a/es-app/src/CollectionSystemManager.cpp
+++ b/es-app/src/CollectionSystemManager.cpp
@@ -198,6 +198,50 @@ bool systemByManufacurerSort(SystemData* sys1, SystemData* sys2)
 	return name1.compare(name2) < 0;
 }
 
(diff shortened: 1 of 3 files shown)


Darknior:
Very good idea to order like it :)

PR #1400, opened 2023-02-03
Add joystick configuration translations in french

Tartifless:
SOUTH, EAST, WEST, NORTH
LEFT and RIGHT SHOULDERS
D-PAD directions
Triggers
Stick pushes


nadenislamarre:
@fabricecaruso as a general rule, .po msgid is an automatic process. adding them manually will remove them. i've fixed in a today commit.

PR #1402, opened 2023-02-08
adding more HotD names for compatibility

Tovarichtch:
Some ROMs use "the" at the start of the name instead at the end. This will fix the name variation.


nadenislamarre:
<game>houseofthedead2</game>
    <game>thehouseofthedead2</game>

the 2nd one is unusefull because if houseofthedead2 is not found in the name, thehouseofthedead2 cannot be found.

PR #1413, opened 2023-03-06
Update Arabic

Chipsum:
100 entries only left :love_you_gesture:


nadenislamarre:
thanks. the file was integrated (not via this pr cause of conflict)
you can check it on translations.batocera.org

PR #1417, opened 2023-03-10
Update Arabic (100%)

Chipsum:
Finished translating.
https://photos.app.goo.gl/HqNLE8HAw1Q1CmPd8


nadenislamarre:
merged. but reverted.
you put the options file here. so it removed all translations.
you should use translations.batocera.org

PR #1459, opened 2023-05-16
Update Font.cpp

Chipsum:


fabricecaruso:
What's the point you just changed a comment ?

Chipsum:
> What's the point you just changed a comment ?

That's the point, I just changed a comment.
The reason behind that is that the new font supports Persian too, which is a language that can be added to BATOCERA.

lbrpdx:
@Chipsum Can you help with the Persian translation? The process is explained on the wiki https://wiki.batocera.org/help_with_translation -- thank you!

Chipsum:
> @Chipsum Can you help with the Persian translation? The process is explained on the wiki https://wiki.batocera.org/help_with_translation -- thank you!

I have no intention of contributing to the BATOCERA project anymore. But I see no reason why Persian should not be there if there are translators ready to contribute.

lbrpdx:
> I have no intention of contributing to the BATOCERA project anymore. But I see no reason why Persian should not be there if there are translators ready to contribute.

Too bad. That's right: we need more translators for the missing languages (and even completing the translations that are not fully up to date). But that's a lot of hard work, that requires motivated people. And they're harder to find than people who just want others to do the work for them 😉

Chipsum:
no it's just that we take so much bull*hit , like fabrice and how shocked he is to see someone trying to add something to a comment, just trying to help sometimes is not good at all.
Closing...

PR #1463, opened 2023-05-16
[Suggestion] web frontend can look ES capability

NullPopPoLab:
add web API /caps
returns ES capability info in JSON

frontend can understand (customised) ES features
and switch correctly.


NullPopPoLab:
I try implementing over proxy
https://github.com/NullPopPoLab/BatoceraProxy

but it works for my custom ES.
as possible, hope to switch for official Batocera versions.

PR #1466, opened 2023-06-02
Fix Atari XE scraping (#1405)

### LINKED ISSUE #1405: xegs games cannot be scraped
For system "xegs", no scrapers can be selected in the ES menus (when ungrouped).

From examining the code, I think the problem is that the platform ID list uses the name "atarixe" while es_systems.cfg uses the platform name "xegs"

https://github.com/batocera-linux/batocera-emulationstation/blob/master/es-app/src/PlatformId.cpp#L26

I believe adding this line would fix it:
```
		{ "xegs",				ATARI_XE },
```

lbrpdx:
Fix #1405


lbrpdx:
From original report:
> It's worth noting that even with this patch, it's necessary to scrape this system from TheGamesDB, as other scrape APIs (including ScreenScraper) don't provide data for Atari XE system.

PR #1470, opened 2023-06-21
Update getArchString() function to match batocera board renaming

rtissera:
Following a board renaming in batocera, this PR updates the getArchString() accordingly.

It tries first to read the /usr/share/batocera/batocera.arch file (only when not on Windows) and fallbacks to classic #ifdef implementation.

### DIFF
--- a/es-core/src/platform.cpp
+++ b/es-core/src/platform.cpp
@@ -420,7 +420,10 @@ std::string getArchString()
 {
 #if WIN32
 	return "windows";
-#endif
+#else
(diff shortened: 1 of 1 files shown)


rtissera:
Ping @fabricecaruso for validation.

fabricecaruso:
/usr/share/batocera/batocera.arch file will always be deployed ?

rtissera:
@fabricecaruso in batocera builds, yes.

PR #1490, opened 2023-07-20
[Scripts] Allow run scripts without waiting for the launcher process to end

fabricecaruso:
- Refactor platform into namespace Utils::Platform then create ProcessStartInfo class that handles options (ie. waitForExit).


nadenislamarre:
i thought that this log :
https://github.com/batocera-linux/batocera-emulationstation/pull/1490/files#diff-04a33ac0834d48df29937051f5ec15ce306ee1d531293af290fbd3fc826e8fa7R133

was only for games outputs.
in case we have an even on game-end, then, we'll have not the game output.
anyway, it is rare.

PR #1502, opened 2023-08-10
add model2 gun only games

dmanlfc:


nadenislamarre:
i think it alread worked.
i think it uses the mamegames.xml file for this one ? guns uses 2 files in fact cause one already existed for mame and arcade games.

PR #1509, opened 2023-08-25
Fix hang when running ES scripts asynchronously

n2qz:
When not waiting for child scripts, reap zombies and avoid calling exit handlers

- Call fork() twice and wait for first child, to prevent zombie processes.
- Call _exit() instead of exit() from the forked subprocesses, to prevent flushing STDIO buffers and skip calling exit handlers.
- Use execl() instead of system() to avoid keeping the intermediate forked grandchild around longer than necessary.

### DIFF
--- a/es-core/src/utils/Platform.cpp
+++ b/es-core/src/utils/Platform.cpp
@@ -5,11 +5,13 @@
 #if WIN32
 #include <codecvt>
 #else
+#include <sys/types.h>
 #include <unistd.h>
 #include <ifaddrs.h>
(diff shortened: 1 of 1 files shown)


nadenislamarre:
it restrics commands to shells

n2qz:
> it restrics commands to shells

There's no restriction, the execl() call is effectively what standard implementations of system() use internally.  Thanks for reviewing and merging!

nadenislamarre:
thanks a lot for the explanation and your help.

PR #1522, opened 2023-09-02
Initial wheel games database (WIP, VERY INCOMPLETE)

Tovarichtch:
Trimming down to only games with wheel support (dial, paddle, analog stick input that have axis-range).


Tovarichtch:
Still missing list: xbox 360 & wiiu

PR #1527, opened 2023-09-23
[New SaveStates] Better savestates management using es_savestates.cfg + Add 2 new options.

fabricecaruso:
- Add SHOW RETROACHIEVEMENTS ICON & SHOW FOLDERS FIRST options
- SaveStates : For autosaves, use "-state_slot <id> -state_file <filename>" arguments instead of "-state_slot autosave"


askotx:
Maybe, I'm wrong, but could it be easier to use retroarch self feature for selecting auto savestate entry slot parameter, like in: https://github.com/libretro/RetroArch/pull/13354

And I repeat, if I'm wrong, sorry

PR #1532, opened 2023-09-28
Use pm-is-supported to check suspend support

n2qz:
Use pm-is-supported to check suspend support, making SUSPEND SYSTEM option available only when appropriate


fabricecaruso:
@nadenislamarre I let you handle this one.

nadenislamarre:
it looks good.

PR #1534, opened 2023-09-30
Adding Wii u and Xbox 360 wheel games

Tovarichtch:


fabricecaruso:
there are conflicting files...

Tovarichtch:
I don't even understand what's conflicting.

PR #1535, opened 2023-10-02
fix crash

teslakang:
fix crash


teslakang:
[2a5088e] : fix crash rare case. (maybe when no rom file exist)

[2d26571]: fix crash when not exist resource directory.

fabricecaruso:
I tried with Korean but can't reproduce the debug assertion with the keyboard...
Except the fact that I can't reproduce this case, I agree to merge your 2 PR, once Batocera 38 is released.

teslakang:
I think this issue only occurs in Korean Windows or non-English Windows environments.

PR #1537, opened 2023-10-02
fix win32 locale processing

teslakang:
fix win32 locale processing

### DIFF
--- a/es-core/src/LocaleES.cpp
+++ b/es-core/src/LocaleES.cpp
@@ -256,8 +256,19 @@ void EsLocale::checkLocalisationLoaded()
 		std::string line;
 
 		std::ifstream file(xmlpath);
-		while (std::getline(file, line))
+		std::string old_line;
+		while (1)
 		{
+			if (old_line.empty())
+			{
+				if (!std::getline(file, line)) break;
+			}
+			else
+			{
+				line = old_line;
+				old_line = "";
+			}
+
 			if (line.find("\"Plural-Forms:") == 0)
 			{
 				auto start = line.find("plural=");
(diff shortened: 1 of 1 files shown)


fabricecaruso:
What is this supposed to fix ? Any exemple ?
 We need more details.

teslakang:
Old code does not support multiline translation.
It has been modified to support it.

fabricecaruso:
Can you provide a sample ? so we can reproduce & test

teslakang:
This only happens on Windows.
If the execution path of the emulationstation contains a folder written in Korean(한글), Japanese(일본), Chinese(중국), etc. rather than ASCII characters, emulationstation will not run properly.
In my case...
"d:\애뮬\emulationstation" is cannot load xml or image...

The reason for this is that emulationstation uses UTF8 internally, but Windows uses multibyte characters.
Therefore, all file input and output must be process through Unicode functions.
Currently, I have worked on images and xml, but there may be some things I don't know, so additional related work may be needed.

teslakang:
EmulationStation Windows build uses gettext by directly implementing it.
So it does not support all specifications.
I don't know much about the rest, but I think not supporting multi-line translation is a big problem.
This is a small patch to solve it.

msgid ""
"THIS WILL DELETE THE ACTUAL GAME FILE(S)!\n"
"ARE YOU SURE?"
msgstr ""
"실제 게임 파일이 삭제됩니다!\n"
"확실합니까?"

That is "locale/lang/ko/LC_MESSAGES/emulationstation2.po" game delete message...

Old code...
image

My code..
image

PR #1552, opened 2023-10-24
Make "a" (EAST) button skippable in controller config gui

n2qz:
Currently, the EAST button is not skippable in the controller config gui.  This leads to an awkward and unpleasant user experience when configuring minimal controllers with few buttons.  In these situations, as a user, I want to be able to skip a mapping for EAST and provide a mapping for SOUTH.

### DIFF
--- a/es-core/src/guis/GuiInputConfig.cpp
+++ b/es-core/src/guis/GuiInputConfig.cpp
@@ -48,7 +48,7 @@ void GuiInputConfig::initInputConfigStructure()
 {
 	GUI_INPUT_CONFIG_LIST =
 	{
-		{ "a",                false, InputConfig::buttonDisplayName("a"),    InputConfig::buttonImage("a") },
+		{ "a",                true,  InputConfig::buttonDisplayName("a"),    InputConfig::buttonImage("a") },
 		{ "b",                true,  InputConfig::buttonDisplayName("b"),    InputConfig::buttonImage("b") },
 		{ "x",                true,  "NORTH",              ":/help/buttons_north.svg" },
 		{ "y",                true,  "WEST",               ":/help/buttons_west.svg" },


nadenislamarre:
@fabricecaruso you confirm it is not something wanted ? (in such case, a comment should be added to be sure)

fabricecaruso:
And how will you ever do something if you can't define / press this button ?
Knowing that you need this button to navigate in emulationstation, and later, play games.

This button is required in this screen & I don't see any good reason to remove the required flag.
If you want to hack a controller for you own usage, then do it manually in es_input.cfg file.

Unless you have a very good reason that convinces us, I don't agree to merge this one.

n2qz:
> And how will you ever do something if you can't define / press this button ? Knowing that you need this button to navigate in emulationstation, and later, play games.

It's possible to navigate the menus with a keyboard or another controller, while using a one-button joystick to play games.

fabricecaruso:
Well. This is understandable. 
But I don't think it's a good move to make everything optionnal.

As I don't have & can't test such a controller :
What reports SDL_JoystickNumButtons ( or  inputConfig->getDeviceNbButtons() ) ? 
If it reports 1, it may be interesting to manage differently.
With 1 only SOUTH would be required. With >= 2 SOUTH and EAST would be required.

Can you indicate the value you have with your joystick ?

n2qz:
It's a virtual controller created by a userland GPIO device driver I'm developing for the Batocera riscv target.  It's currently used with a standard Atari 2600 joystick, presenting as five buttons mapped in ES to up, down, left, right, and SOUTH.

```
[root@jack /userdata/system]# sdl2-jstest --list
Found 1 joystick(s)

Joystick Name:     'gpio2uinput controller'
Joystick Path:     '/dev/input/event4'
Joystick GUID:     000000006770696f3275696e70757400
Joystick Number:    0
Number of Axes:     0
Number of Buttons:  5
Number of Hats:     0
Number of Balls:    0
GameControllerConfig:
  missing (see 'gamecontrollerdb.txt' or SDL_GAMECONTROLLERCONFIG)
Button code  0:   304
Button code  1:   305
Button code  2:   306
Button code  3:   307
Button code  4:   308
```

fabricecaruso:
Here's another suggestion.

I don't understand why SOUTH is historically skippable, as it's the main & OK button with every platform.
Looks like we inherited this behaviour from Retropie.
As it's the default, it should be shown first & it should not be skippable.

Now, about EAST :

The skippable flag for EAST should depend on the button count, so : 
- We must pass the inputconfig to the initInputConfigStructure method.
- Refactor the GUI_INPUT_CONFIG_LIST to make SOUTH always first.
- Change skippable for SOUTH & EAST to be non skippable by default.
- Switch name (button internal name) with INVERTEDINPUTCONFIG ( this is for Windows ) -> This makes SOUTH always 1st on both plaforms.
- Make "EAST" ( 2nd item ) skippable only if the button count is 1 or if it's 5 with no axis & no hats.

What do you think about this ?

image

fabricecaruso:
So what shall we do ? Are you going to change your PR, or should I close it and add what I suggested ?

n2qz:
I don't think I could run with your suggestion and get it right, so I'm happy to let you take on the implementation if you're willing.

fabricecaruso:
Ok. I merged this one : https://github.com/batocera-linux/batocera-emulationstation/pull/1567

PR #1560, opened 2023-10-30
Select running board as default installation target

n2qz:
A common source of user error when using the "INSTALL ON A NEW DISK"
    feature is that the user guesses their architecture incorrectly.

    With this change, if an installation architecture matches the
    currently running board, it will be selected as default.

    If no match is found, the previous behavior is retained.

### DIFF
--- a/es-app/src/ApiSystem.cpp
+++ b/es-app/src/ApiSystem.cpp
@@ -1961,6 +1961,15 @@ std::string ApiSystem::getRunningArchitecture()
 	return "";
 }
 
+std::string ApiSystem::getRunningBoard()
+{
+	auto res = executeEnumerationScript("cat /boot/boot/batocera.board");
+	if (res.size() > 0)
+		return res[0];
+
+	return "";
+}
+
 std::string ApiSystem::getHostsName()
 {
 	auto hostName = SystemConf::getInstance()->get("system.hostname");
--- a/es-app/src/ApiSystem.h
+++ b/es-app/src/ApiSystem.h
@@ -228,6 +228,7 @@ public:
 	virtual std::vector<std::string> extractPdfImages(const std::string& fileName, int pageIndex = -1, int pageCount = 1, int quality = 0);
 
 	virtual std::string getRunningArchitecture();
+	virtual std::string getRunningBoard();
 
 	std::vector<PacmanPackage> getBatoceraStorePackages();
 	std::pair<std::string, int> installBatoceraStorePackage(std::string name, const std::function<void(const std::string)>& func = nullptr);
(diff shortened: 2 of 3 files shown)


n2qz:
Related issue in batocera.linux repository:  https://github.com/batocera-linux/batocera.linux/issues/7975

fabricecaruso:
I'm OK with this one. @nadenislamarre I let you handle this one ?

nadenislamarre:
hum. next time people will complain the default is wrong and badly oriented them.
i'm not against this change ; i would like the mind of other people of the team.

Hew-ux:
Being in the community often, I see 99.99% of users are using this option to install Bato onto the same machine, but to its "physically inaccessible" internal drive. This change will not cause issues, it will solve them.

dmanlfc:
Great idea @nadenislamarre we should do it.

n2qz:
Found a bug, placing this in draft status until fixed

nadenislamarre:
hum ok.

nadenislamarre:
@fabricecaruso for me the pr is ok to validate.
note that the api is not protected by a ifdef nor by a white list script, but while it is only called from the installer, it is ok for retrobat i think. Can you merge if that's ok for you on this side ?

fabricecaruso:
What's the difference between :

/boot/boot/batocera.board -> What you use in the new api
/usr/share/batocera/batocera.arch -> What already exists in Platform::getArchString() 

Looks like it's 2 different fiches for the same thing, no ?
Isn't it a duplicate ?

n2qz:
> What's the difference between :
> 
> /boot/boot/batocera.board -> What you use in the new api /usr/share/batocera/batocera.arch -> What already exists in Platform::getArchString()
> 
> Looks like it's 2 different fiches for the same thing, no ? Isn't it a duplicate ?

They are the same for boards where there are no subtargets, but different for boards where there are multiple subtargets.  

`/boot/boot/batocera.board` -> the subtarget name for a board build (`BATOCERA_SUBTARGET`)
`/usr/share/batocera/batocera.arch` -> A value derived from the target named in `BR2_PACKAGE_BATOCERA_TARGET_*` in `package/batocera/core/batocera-system/batocera-system.mk`

For example, there are different installation target image files for e.g. ps5000 vs ps7000 corresponding with different `/boot/boot/batocera.board` while they share the same `/usr/share/batocera/batocera.arch` containing "rk3128".  So, I believe batocera.board contains the right name to use in this context.

nadenislamarre:
- we built batocera for a arch (let's say s922x) - 4 hours
- then, from the build, we create several images for different boards : odroidn2, gtkingpro, ogu, vim3, ... - 5 minutes by image

we do that to win a lot of compilation time.
all but /boot, and then, /usr/share/batocera/batocera.arch is then part of the compilation. It cannot differ for the different s922x boards. it contains s922x. We cannot create /usr/share/batocera/batocera.board, because the file is the same for all boards.

it seems we need sometimes the real name of the board (if think for example for custom audio output, screen config, ...)
thus we added in /boot the name of the board.

n2qz:
> it seems we need sometimes the real name of the board (if think for example for custom audio output, screen config, ...) thus we added in /boot the name of the board.

I guess we agree then that `/boot/boot/batocera.board` is the correct choice here.

Can this PR be merged?

PR #1631, opened 2023-12-30
integrate guns/wheels db checks

nadenislamarre:


Tovarichtch:
This is totally outdated. Still needed?

PR #1638, opened 2024-01-02
Optimizing wheel (part 3)

Tovarichtch:
`controller="1"` means a controller must be set to port 1
`port1="2"` means the wheel (on port 1) must be set to port 2
Default N64 and Saturn mapping need to be redone.


Tovarichtch:
I'm remaking this. I fucking hate it when I'm merging an update, it fucks everything up.

PR #1648, opened 2024-01-09
add-metadatagames.xml

Mikhailzrick:
To be used with batocera updated code for handling generic metadata for games


fabricecaruso:
What's the use of it as it will not be used by emulationstation's code ?

Mikhailzrick:
> What's the use of it as it will not be used by emulationstation's code ?

It's going to be used in a similar way as wheelgames.xml and gungames.xml which are all in the recourses directory. This file is just going to be for more generic games and systems.

fabricecaruso:
I still don't really see the interest of adding this in emulationstation repository, but, if we do :

- I'm not convinced it's a good thing to add new icons (in the game name) for about everything expecially this kind of things which is very system specific.
As metadatas it could eventually be interesting, but most themes won't ever display that information.
Also, these metadatas could be exposed only if described precisely, which will probably won't be easy as it's very system specific.
How themers will know what to use ?

- Your file format is different that the one in gungames.xml and wheelgames.xml. If we have to load your file, we have to write another new piece of code to parse almost the same thing, as element are different. 
It would be far better if it was the exact same conventions for elements :
  -> replacing `<n64>` element with `<system name="n64"> `
  -> replacing `<game name="007theworldisnotenough" widescreen="true" rumble_pak="true" />` with `<game widescreen="true" rumble_pak="true">007theworldisnotenough</game>`

- It's again a new file... It would probably be better if we merge everything into one single file. (gungames.xml, wheelgames.xml, and your file )

Mikhailzrick:
The file format and structure used was based on Susan’s recommendation. I don’t know how we want to handle merging the files together or what eventually format will be used. I suspect it’ll be a work in progress especially as wheels, guns, and generic games will actively be updated. I plan to systematically go through the libretro cores first and add games that support native widescreen to be used to auto switch to 16:9 aspect ratio on game launch.

Tovarichtch:
I agree with one central metadata file having all the specifics. Easier to maintain in the long run. We see the need since gungames.xml and wheelgames.xml. Even mamenames.xml has its own parameters for wheels/guns.

However I don't know if a .xml is the best option and most optimize for that. But we do need a main database for all specific settings to run them out of the box.

Darknior:
> However I don't know if a .xml is the best option and most optimize for that. But we do need a main database for all specific settings to run them out of the box.

For sure, and with this ton of games, i think writing that an option is false is useless.
Why writing : widescreen="false"
Only true options or parameters are important. It will be easier to maintain. 
On N64 there will now have widescreen games, rumble games, Wheel games ....
But maybe i'm false ...

> * I'm not convinced it's a good thing to add new icons (in the game name) for about everything expecially this kind of things which is very system specific.
>   As metadatas it could eventually be interesting, but most themes won't ever display that information.
>   Also, these metadatas could be exposed only if described precisely, which will probably won't be easy as it's very system specific.
>   How themers will know what to use ?

I'm divided on this point, if we have too many icons after the name the list will become unreadable...
Whether it's the Rumble or the Widescreen, it's better to display them elsewhere, even if they are important features.
After the themes display them or not, it only depends on the creators... as long as the available icons (metadata) are listed in the wiki.

Mikhailzrick:
> > However I don't know if a .xml is the best option and most optimize for that. But we do need a main database for all specific settings to run them out of the box.
> 
> For sure, and with this ton of games, i think writing that an option is false is useless. Why writing : widescreen="false" Only true options or parameters are important. It will be easier to maintain. On N64 there will now have widescreen games, rumble games, Wheel games .... But maybe i'm false ...
> 
> > * I'm not convinced it's a good thing to add new icons (in the game name) for about everything expecially this kind of things which is very system specific.
> >   As metadatas it could eventually be interesting, but most themes won't ever display that information.
> >   Also, these metadatas could be exposed only if described precisely, which will probably won't be easy as it's very system specific.
> >   How themers will know what to use ?
> 
> I'm divided on this point, if we have too many icons after the name the list will become unreadable... Whether it's the Rumble or the Widescreen, it's better to display them elsewhere, even if they are important features. After the themes display them or not, it only depends on the creators... as long as the available icons (metadata) are listed in the wiki.

As far as the icons go, I was just saying if the data is there, theme makers may be able to use it if they choose. It’s not something I plan on working on.

as far as the false flags, you may be right, it may be much more beneficial to only include “true” flags as right now I can’t really think of a use case for what I have planned anyway for needing to know a “false”. I’ll probably update it soon and remove all the false attributes.

fabricecaruso:
mamenames.xml / mamebioses /  mamedevices can't be merged with other xml files and should stay appart because :
- They are generated files : I have a c# piece code that generate these files from the huge xml mame files (the latest is about 260mb). Every specific information (gun/vert) are imported from the mame Xml. My piece of code can be modified to extract new information one day if needed.
- It does not use the same method for searching games. Arcade games roms have normalized names, so it's easier to detect than the other systems, where there is no real convention.

For this reason, EmulationStation will probably continue to use mamenames.xml for arcade games.

The question is about the "non-arcade" systems like n64. 

We first created gungames.xml to reference the games supporting the guns for non arcade games.
Then, @nadenislamarre created wheelgames.xml in order to detect games supporting steering wheels.

We already have many files to load. Adding new files is not a good idea, when it's based on the same structure and principles (only some attributes change).
So it's probably time to rationalize this into a unique "database". 

I continue to think the xml format is still the best option. Creating a real database (like sqlite) would require to add new dependencies to emulationstation (c++), read them in python for batocera, read them in c# for retrobat, so it's complex to code, then it will be complex to maintain and create small PRs with small changes as it will be binary files. 
Changing the format ( to json or some text format ) would force to rewrite everything too. 

Even if XML is very verbous, it's still the best option to keep it readable and editable for humans, as we can't automate the building of this database from another big database like we can do with mamenames.

So @Mikhailzrick, I suggest you reformat the XML in this PR to respect the format of gungames.xml / wheelgames.xml like I suggested sooner.
Once done, it will be easy to take the 3 files and merge them into one single unique file. 
I - or somebody else - can easily write code to automate this, or even do it manually.

Then, we'll adapt our respective codes to read that file (EmulationStation / configgen for batocera / emulatorLauncher for retrobat ).

Are you OK ?

Mikhailzrick:
> mamenames.xml / mamebioses / mamedevices can't be merged with other xml files and should stay appart because :
> 
> * They are generated files : I have a c# piece code that generate these files from the huge xml mame files (the latest is about 260mb). Every specific information (gun/vert) are imported from the mame Xml. My piece of code can be modified to extract new information one day if needed.
> * It does not use the same method for searching games. Arcade games roms have normalized names, so it's easier to detect than the other systems, where there is no real convention.
> 
> For this reason, EmulationStation will probably continue to use mamenames.xml for arcade games.
> 
> The question is about the "non-arcade" systems like n64. First of all, roms names are not normalized for other system, and there is no global database we could import, so we did the kind of encoding for names, like you see in the current xml files.
> 
> We first created gungames.xml to reference the games supporting the guns for non arcade games. Then, @nadenislamarre created wheelgames.xml in order to detect games supporting steering wheels.
> 
> We already have many files to load. Adding new files is not a good idea, when it's based on the same structure and principles (only some attributes change). So it's probably time to rationalize this into a unique "database".
> 
> I continue to think the xml format is still the best option. Creating a real database (like sqlite) would require to add new dependencies to emulationstation (c++), read them in python for batocera, read them in c# for retrobat, so it has a complex impact on multiple projects. Also, it will be complex to maintain and create small PRs with small changes as a real database is based on binary files. Changing the format ( to json or some text format ) would force to rewrite everything too.
> 
> Even if XML is very verbous, it's still the best option to keep it readable and editable for humans, as we can't automate the building of this database from another big database like we can do with mamenames.
> 
> So @Mikhailzrick, I suggest you reformat the XML in this PR to respect the format of gungames.xml / wheelgames.xml like I suggested sooner. Once done, it will be easy to take the 3 files and merge them into one single unique file. I - or somebody else - can easily write code to automate this, or even do it manually.
> 
> Then, we'll adapt our respective codes to read that file (EmulationStation / configgen for batocera / emulatorLauncher for retrobat ).
> 
> Are you OK ?

Agreed for now xml is the way to go. I’ll reformat the file in my PR to match what’s in guns and wheels, there apparently was a misunderstanding somewhere when Susan was advising me how to format this new generic metadata file. But I agree on all subjects. I’ll fix soon and push

Mikhailzrick:
@fabricecaruso xml file format updated and fixed.

fabricecaruso:
Well... I just wrote a piece of code to merge the 3 files, and the results need reflection.
Here are samples of what it would give :

- using a simple merged file : [merged _basic.zip](https://github.com/batocera-linux/batocera-emulationstation/files/13891523/merged._basic.zip)
The gun attribute is already used in gungams, so if we create a gun="true" attribute, we have a conflict. So we'd have to rename gun to something like gun_type.
The same happens with wheelgames and the wheel attribute.
Using this file, it would be easy to adapt codes, but it's barely readable when there are many attributes, and it will be probably complex to maintain if we add new attributes.

- using a more complex merge : [merged_changed.zip](https://github.com/batocera-linux/batocera-emulationstation/files/13891548/merged_changed.zip)
The game is a gungame if we have the attribute game="true" OR if we have a child element <gun> with gun attributes ( type, .. )
The game is a wheel game if we have the attribute wheel="true" OR if we have a child element <wheel> with wheel attributes ( type, rotation, deadzone... ). 
We could avoir the game=true attribute and write an empty <gun/> or <wheel/> element, insead.
New metadatas would have it's own child element <metadata>
Doing something like this is probably more maintenable on the long term, but it required more work to adapt.

What do you all think ?

Darknior:
For my part the second solution is better. Easier to read too.

```
    <game name="perfectdark">
      <metadata widescreen="true" rumble_pak="true" />
    </game>

    <game name="burnout2">
      <wheel wheel_type="GTForce" wheel_rotation="200" />
    </game>

    <game name="endgame">
      <gun gun1port="2" />
    </game>
```

Just why rumble_pak and widescreen are in METADATA ?
Because maybe widescreen will ask for some more parameters for some games ...
Like launching a cheat code in libretro to activate the widescreen in game options.

And for sure some other options can come, like an auto 6 butons parameters for the Megadrive games that are better to play with a 6 button pad hat is not by default ... etc etc ...
All these options will be METADATA ?

In non Metadata we will maybe have games that use Micro !! for exemple, that is an accessory too.

And some games in arcade i know use WHEEL and GUN at same time, but this method let it possible :)

Tovarichtch:
Hard for me to tell. Basic format is a lot simplier and faster for me to work on (and I'm pretty much the only one since a long time doing so), but the second one has the merit to be clearer in the first look, but the user **must** know what it means. I can see why the second format can look appealing, but let's keep in mind that not many people are going to work and maintain those lists.

Let's demonstrate with one complex use case : Starsky & Hutch on PS2 (not on the list yet for compatibility reason). This game is *at the same time* a driving game (using GT Force wheel which is limited to 200 degree rotation angle and has a tiny deadzone) and a gun game (using GunCon 2) but has no 16:9 native screen. Wheel always needs to be on port 1, but gun must be on port 2.

Therefore, in the first example we write the line like this : 

```
    <game name="starskyandhutch" gun="true" gun1port="2" wheel="true" wheel_type="GTForce" wheel_rotation_angle="200" />
```

A bit more code, but it's one line. Shorter.

Now in the second format : 

```
    <game name="starskyandhutch">
      <gun gun1port="2" />
      <wheel wheel_type="DrivingForcePro" wheel_rotation_angle="200" wheel_deadzone="10" />
    </game>
```

More clear, bit longer and more lines. But this is more similar to mamenames.xml. We know that this game is gun compatible and wheel compatible by a quick glance and everything else on the right is an optimized setting. Having worked so damn much on both lists, I think the second format is better but not in term of speed writing (for me at least). So that when starting with `<gun [...]>` we know right away it's `game gun="true"`.

Last thing is that `metadata` is a bit ugly and condensing. We could use it as a starting point, but not in the long run. I prefer to separate it, in the spirit of my previous example above : 

```
    <game name="perfectdark">
      <video widescreen="true" />
      <controller rumble="true" />
    </game>
```

In this case, we can have a standard "video" parameters for game *about video settings* (widescreen, color palette, shader, etc.) that fixes and accurately make the game playable just like the hardware. Same for settings *about controller* like rumble, analog, accessory specific (camera, microphone, drum, etc.) required by games to be playable. Light guns and steering wheels are too important with too many technicalities to be integrated to `<controller [...] />`. Same thing for settings *about sound* `<sound audio_latency="64" mono="true" />` as an example. It can happens for few games. Not all games are the same, keep that in mind and having a standard structure like this could help maintain the list.

With all that said, I aim for accuracy so this list must avoid hacks and personal taste like upscaling. If a game has features that we can use it, we must use it whenever possible. Giving users the choice to enable custom settings (video, cheat, controller feature, sound, etc.) but keeping the default setting as accurate as possible. That's why I've recently changed the gungames.xml for PSX using the `gun_type="justifier"` as default because the games only supported that type, but couldn't up to recently because there was no known working core supporting that type of gun. PCSX-reARMed now does, therefore I ditched out the GCon45 support (possible with ROM patch) to native support (impossible with the ROM patch).

Mikhailzrick:
Just to add my thoughts about, my intended use case for this file is going to basically involve being able to parse it and find out information about specific games and what they supported or native features they had. The examples above about being able to determine if a game natively supported anamorphic widescreen, or was compatible with the n64 rumble pack is a good example, but there could be a lot more attributes when I start working on other systems. So my vote at this time would be for whichever option is the easiest to modify and add in currently unknown attributes. So if I understand it all correctly I think that would be option 1. 

I’m not sure about parsing speed, but also having fewer nested elements may be faster?

Darknior:
For my part i stay on the second one, it is easier to read, to understand, to update.

```
    <game name="perfectdark">
      <video widescreen="true" />
      <controller rumble="true" />
    </game>
```
    
This is a good exemple :)

It is same time to parse, you will nit see any difference.
And if you prefer one line format you can write it like that, it's same

`<game name="perfectdark"><video widescreen="true" /><controller rumble="true" /></game>`

With time the first one, faster to write yes, will be hardest to maintain, and you will regret your choice @Tovarichtch  ...
For sure one unique game will never have all the options, but when you will search something in a big big file, with more than 1000 games ... it will be a pain in first format :(

Tovarichtch:
> With time the first one, faster to write yes, will be hardest to maintain, and you will regret your choice @Tovarichtch ... For sure one unique game will never have all the options, but when you will search something in a big big file, with more than 1000 games ... it will be a pain in first format :(

Well, *right now*, I'm maintaining both wheel and gun lists alone. I've put ton of hours in this and I still will continue. You are right on one thing tho: futureproof. If we want new contributors, we must make it easy to understand. Maybe one day I'll stop contributing to the project and move on, the legacy should be understandable.

Mikhailzrick:
Yeah, I think I agree on that point. For future maintainability it might be better to go that route.

fabricecaruso:
It's a bit verbose, but here's a new try with suggestions about video/controllers :

[merged.zip](https://github.com/batocera-linux/batocera-emulationstation/files/13907656/merged.zip)

Tovarichtch:
Seems like this is the best compromise, for now. Looks heavy to my eyes but crystal clear. Hard to argue against this format.

Mikhailzrick:
How do we want to go about this then? Merge this PR that follow the same structure as the other wheels and guns file, then do a separate PR to merge them?

Tovarichtch:
I suggest we work on this and merge for v40 instead, v39 is close to RC1.

Mikhailzrick:
I have a PR that’s dependent on this file, can we merge for v39 so my other PR can be merged, then work on merging the xml files and updating the controllersConfig.py script for v40?

fabricecaruso:
In a first time, the best solution is probably to let both databases cohabit ( keeping old format files, an the one with new format in the repository ) to let time for everyone to adapt.

Once everyone is ready, we'll delete the old files.
Are you OK with it ?

fabricecaruso:
I merged your file, and I just added the "gamesdb.xml" file into the resources in this PR 
https://github.com/batocera-linux/batocera-emulationstation/pull/1652

Contact me when you're OK on configgen's side, and the old files will be removed.

fabricecaruso:
Finally that was fast to code in EmulationStation. See https://github.com/batocera-linux/batocera-emulationstation/pull/1653

Tovarichtch:
Well that was quick. Should I update my PR to include the changes in this new file too?

fabricecaruso:
Please

fabricecaruso:
I just finished the implementation on retrobat's side. See https://github.com/RetroBat-Official/emulatorlauncher/pull/282

Tovarichtch:
> Collaborator

This is going to take me a crazy amount of time, but will do.

Mikhailzrick:
@fabricecaruso I’ve updated what I needed to use the new gamesdb.xml file. So I actually don’t even need the metadata games.xml file.

fabricecaruso:
Good. Do configgen still use wheelgames.xml and gungames.xml ?

Tovarichtch:
Yes. @nadenislamarre will need to update it

PR #1652, opened 2024-01-12
[GamesDB] Initial version for gamesdb.xml file, which is a merge of wheelgames.xml, gungames.xml and metadatagames.xml

fabricecaruso:
These 3 files will be removed once every launcher is adapted.

### DIFF
--- a/resources/gamesdb.xml
+++ b/resources/gamesdb.xml
@@ -1,0 +1,3298 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<systems>
+  <system name="3do">
+    <game name="drugwars">
+      <gun />
+    </game>
+    <game name="lostgold">
+      <gun />
+    </game>
+    <game name="maddogii">
+      <gun />
+    </game>
+    <game name="crimepatrol">
+      <gun />
+    </game>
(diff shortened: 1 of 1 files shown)


Tovarichtch:
Just letting you know that some mistakes were made in wheelgames.xml when you created the file. It's not `wheel_rotation` but `wheel_rotation_angle` that is the correct metadata. Plus, I've done many changes in my last draft PR for steering wheel games.

fabricecaruso:
I don't understand. 

Where did you see any wheel_rotation_angle attribute in https://github.com/batocera-linux/batocera-emulationstation/blob/master/resources/wheelgames.xml ?

Anyway : 
- wheelgames shall be deleted soon.
- In the new gamesdb file, it was renamed to "rotation" in the wheel element.
- I added a xsd schema file for gamesdb. It will be better for everyone to respect conventions with a schema. We now all shall respect the schema, and every evolution will require the schema to be updated too.

Tovarichtch:
This is what I'm saying. I've done a mistake in my previous pr. The correct metadata parameter is `wheel_rotation_angle` ( fixed in #1655 ) **not** `wheel_rotation`.

PR #1657, opened 2024-01-14
[SystemView] Add ability to use imagegrid or textlist to replace carousel in the systemview

fabricecaruso:
image


Darknior:
une dinguerie !!!!

PR #1679, opened 2024-02-03
Add STEAM as PC Windows in ScreenScraper.cpp

toccata10:


fabricecaruso:
Did you really try to compile this ???? 
This will not compile !

toccata10:
Maybe I'm doing something wrong (still have much to learn), but it compiles on my system.
What could be wrong with this one line code addition ?

Screenshot from 2024-02-04 09-12-15
Screenshot from 2024-02-04 09-13-01

fabricecaruso:
What you did :

- Will not compile because STEAM is not defined as an enum value ( PlatformId - see PlatformId.h ). I don't know how you compiled this, probably compiling using a version of sources that don't include your changes.
- Is useless if you only assign an id to "STEAM" for screenscraper : nothing changed in the declarations that allow to detect the plaformId based on the plaform strings from es_systems.cfg ( see PlatformId.cpp ). So anyway, it will never find it.
- It seems useless too because... Why set STEAM as a platform for a system if the scrapping platform will always be "windows" ? It has interest to use a specific platform and id only if some other scrapers use a separate platformId. As far as I know, it's not the case, so there is no need for this...

Before you propose a PR : Make sure you compile it and make sure you TESTED it before and that it work on your machine - At least with the plaform (linux in your case) you're working with.

toccata10:
I was trying to solve this kind of problem (game is https://store.steampowered.com/app/1342260/SAMURAI_SHODOWN/): the scrapper returns way too many results which are not relevant at all (example here are games exclusive to xbox360 or neo geo pocket), so it's a mess. 
screenshot-2024 02 05-23h12 00
screenshot-2024 02 05-23h12 11
The paradox is it doesn't offer to scrape from the correct game which is https://www.screenscraper.fr/gameinfos.php?plateforme=138&gameid=335284
Unfortunately, the solution is not as easy as the quick fix I imagined. Having uploaded over 300 medias on screenscraper, I was hoping to benefit from them.
So, I'm closing this. Hopefully someone, one day, will improve that.

fabricecaruso:
As I said, and related to the url you provided, the game you are looking for is using platform 138 which is "platform: windows" in es_systems.

@nadenislamarre The platform should not be "pc" ( which is ms-dos - id 135 ) but "windows" ( id 138 ). 
This should be changed here :
https://github.com/batocera-linux/batocera.linux/blob/master/package/batocera/emulationstation/batocera-es-system/es_systems.yml#L2704
Here too :
https://github.com/batocera-linux/batocera.linux/blob/master/package/batocera/emulationstation/batocera-es-system/es_systems.yml#L2299

PR #1682, opened 2024-02-07
[v40] Add settings for disabling BT and PS3 controllers

DavisNT:
Add settings "ENABLE BLUETOOTH" and "ENABLE PS3 CONTROLLER SUPPORT" to "CONTROLLER & BLUETOOTH SETTINGS".

This requires https://github.com/batocera-linux/batocera.linux/pull/11003

### DIFF
--- a/es-app/src/ApiSystem.cpp
+++ b/es-app/src/ApiSystem.cpp
@@ -490,6 +490,16 @@ std::string ApiSystem::getIpAdress()
 	return result;
 }
 
+bool ApiSystem::enableBluetooth()
+{
+	return executeScript("batocera-bluetooth enable 2>&1 >/dev/null");
+}
+
+bool ApiSystem::disableBluetooth()
+{
+	return executeScript("batocera-bluetooth disable");
+}
+
 void ApiSystem::startBluetoothLiveDevices(const std::function<void(const std::string)>& func)
 {
 	executeScript("batocera-bluetooth live_devices", func);
--- a/es-app/src/ApiSystem.h
+++ b/es-app/src/ApiSystem.h
@@ -167,6 +167,8 @@ public:
 
 	virtual std::string getIpAdress();
 
+	bool enableBluetooth();
+	bool disableBluetooth();
 	void startBluetoothLiveDevices(const std::function<void(const std::string)>& func);
 	void stopBluetoothLiveDevices();
 	bool pairBluetoothDevice(const std::string& deviceName);
--- a/es-app/src/guis/GuiControllersSettings.cpp
+++ b/es-app/src/guis/GuiControllersSettings.cpp
@@ -89,6 +89,48 @@ GuiControllersSettings::GuiControllersSettings(Window* wnd, int autoSel) : GuiSe
 	{
 		addGroup(_("BLUETOOTH"));
 
+#if defined(BATOCERA)
+		// Bluetooth enable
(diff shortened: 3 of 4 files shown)


lbrpdx:
Why would we need a specific case for PS3 controller support? (I mean compared to the dozens of other BT controllers supported)

DavisNT:
@lbrpdx Unfortunately PS3 controller support also enables CVE-2023-45866 vulnerability (source: "Mitigation" section of https://ubuntu.com/security/CVE-2023-45866 ). 

Currently the `ClassicBondedOnly` setting in `/etc/bluetooth/input.conf` is set according to the value of `controllers.ps3.enabled` by [/etc/init.d/S32bluetooth](https://github.com/batocera-linux/batocera.linux/blob/master/package/batocera/core/batocera-bluetooth/S32bluetooth.template). I.e. disabling the PS3 controller support actually fixes CVE-2023-45866 in Batocera (and enabling PS3 controller support enables the vulnerability).

fabricecaruso:
Given the fact that the PS3 thing is mainly related to some kind of vulnerablity, 
I'm not convinced it's a good thing to add this to the main controller menu. User's won't really know why it's there.

You should probably move this option to the developer submenu.

DavisNT:
@fabricecaruso The _ENABLE PS3 CONTROLLER SUPPORT_ setting has description "Might have negative impact on security.", it is in controller settings with the idea that users who don't have PS3 controllers, care about security and have opened _CONTROLLER & BLUETOOTH SETTINGS_ will likely disable it. 
I would like to leave the setting in _CONTROLLER & BLUETOOTH SETTINGS_ unless moving is absolutely necessary.

fabricecaruso:
I would agree if this was for another reason than security, but it's not.
There is no point in polluting the controllers menu when it's all related to a security thing with BlueZ and some HID devices.
We already have users claiming we already have too many options...
We should keep the common menus as simple as possible.

In conclusion, the best place for the "ENABLE PS3 CONTROLLER SUPPORT" option is in the developer options, not in the controller menu. See the "ENABLE HID JOYSTICK DRIVERS" for Windows that is in the developer menu. It's comparable.

So I insist with the fact that it should be moved to developer options.

@lbrpdx Agree ?

lbrpdx:
First of all, if it’s a BlueZ vulnerability, this should be addressed by a bluez update at the OS level.

If users are concerned about security, I believe they are technical enough to edit batocera.conf without an entry in the ES menus. I also agree that we need to keep the menus as simple as possible, they’re already reported as convoluted sometimes. 

IMO we should add the BT enable/disable menu but not the PS3 option.

DavisNT:
@lbrpdx BlueZ has already been updated. Even with the BlueZ update setting `ClassicBondedOnly=false` will enable the CVE-2023-45866 vulnerability.

fabricecaruso:
I personnally have no objection to adding a menu entry for this security thing, as long as it is in the developer options.

DavisNT:
@fabricecaruso Should I then move _ENABLE PS3 CONTROLLER SUPPORT_ to _SYSTEM SETTINGS_ - _FRONTEND DEVELOPER OPTIONS_ immediately below _GUN MOVE TOLERANCE_?

fabricecaruso:
I think it's a good place, yes

DavisNT:
@fabricecaruso Moved the setting and rebased.

PR #1687, opened 2024-02-11
[v40] Save only changed config

DavisNT:
Currently whenever a `batocera.conf` setting is changed in ES all `batocera.conf` settings **modified outside ES** are reverted to the values they had when ES was started.

This PR prevents overwriting all config in `batocera.conf` wheneven one or more config values are changed in ES. With this PR ES updates in `batocera.conf` only those settings that were changed in ES.

Although this is a bugfix I think it would be better to test it for a while (to review and merge it when development of Batocera v40 will start).

### DIFF
--- a/es-core/src/SystemConf.cpp
+++ b/es-core/src/SystemConf.cpp
@@ -69,7 +69,7 @@ bool SystemConf::loadSystemConf()
 	if (mSystemConfFile.empty())
 		return true;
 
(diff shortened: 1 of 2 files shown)


lbrpdx:
LGTM, and I agree this should be merged as we now enter the B40 cycle. Anyone disagrees?

Issue #1686, opened 2024-02-10
Hard Region Detection favors first hard region in filename which may not be desired


geogolem:
### Concerned project

Batocera

### Distribution Version

V39

### Architecture

X86_64

### Issue description

https://github.com/batocera-linux/batocera-emulationstation/commit/d8ba2321e3f58611873128426ae2d95dfb67202d introduced hard region detection.. which is great; however, what happens if the filename has multiple hard regions? Right now, the first hard region in the filename is used, but this might not always be desired.

Sometimes the NoIntro naming convention has multiple regions listed in the filename.

For example on N64:
1080 Snowboarding (Japan, USA) (En,Ja)

or on Master System:
Astro Warrior (Japan, USA, Brazil) (En)
Great Ice Hockey (Japan, USA) (En)
Space Harrier (Japan, USA) (En)

Please see the comment here: https://github.com/batocera-linux/batocera-emulationstation/commit/d8ba2321e3f58611873128426ae2d95dfb67202d#r138488462

Personally I would prefer the region for all of those ROM's to be set to USA instead of Japan... That may not be the case for everybody.

Not sure if this really qualifies as a bug... or a feature..

fabricecaruso:
See my comment here : https://github.com/batocera-linux/batocera-emulationstation/issues/1637#issuecomment-1936945418

geogolem:
> See my comment here : [#1637 (comment)](https://github.com/batocera-linux/batocera-emulationstation/issues/1637#issuecomment-1936945418)

I totally understand your perspective...  I doubt anybody can write a perfect algorithm to handle any rom naming and honestly at this point what you have now with the hardregion detection etc. is near perfect from my perspective.

If you want to keep this issue open and assign it to me, I will try to submit a small PR in the near future... I just wanted to open an issue and get an understanding for how you feel about how it is now and any potential changes. or if i misunderstood or overlooked something...

I think as it is right now --> your algorithm will *mostly* handle NoIntro region/language naming conventions...
https://wiki.no-intro.org/index.php?title=Naming_Convention#Region
https://wiki.no-intro.org/index.php?title=Naming_Convention#Language(s)

It seems issues only arise when there are multiple regions specified.. and even then the current implementation is not strictly incorrect --> it just might not be the "preferred" hard region for any given user/setup. 

[...]

PR #1689, opened 2024-02-15
Region detection improvements

geogolem:
See the discussion here: 
https://github.com/batocera-linux/batocera-emulationstation/issues/1686

### DIFF
--- a/es-app/src/LangParser.cpp
+++ b/es-app/src/LangParser.cpp
@@ -34,12 +34,10 @@ std::string getSystemLang(std::string locale)
 {
 	if (locale.empty())
 		return "en";
-	else
-	{
-		auto shortNameDivider = locale.find("_");
-		if (shortNameDivider != std::string::npos)
-			return Utils::String::toLower(locale.substr(0, shortNameDivider));
-	}
+	auto shortNameDivider = locale.find("_");
+	if (shortNameDivider != std::string::npos)
+		return Utils::String::toLower(locale.substr(0, shortNameDivider));
+	return std::string();
 }
 
 std::string getSystemRegion(std::string locale)
@@ -49,12 +47,14 @@ std::string getSystemRegion(std::string locale)
 	{
 		auto shortNameDivider = locale.find("_");
 		if (shortNameDivider != std::string::npos)
-			region =  Utils::String::toLower(locale.substr(shortNameDivider));
+			region = Utils::String::toLower(locale.substr(shortNameDivider).substr(1));
+
 	}
 	if (region == "gb" || 
 		region == "pt" || 
 		region == "es")
 		region = "eu";
+
 	return region;
 }
 
--- a/es-app/src/LangParser.cpp
+++ b/es-app/src/LangParser.cpp
@@ -50,11 +50,6 @@ std::string getSystemRegion(std::string locale)
 			region = Utils::String::toLower(locale.substr(shortNameDivider).substr(1));
 
 	}
-	if (region == "gb" || 
-		region == "pt" || 
-		region == "es")
-		region = "eu";
-
 	return region;
 }
 
@@ -67,7 +62,7 @@ void LangInfo::extractLang(std::string val)
 		{ { "europe", "eu", "e", "ue", "euro" }, "", "eu" },
 
 		{ { "w", "wor", "world" }, "en", "wr" },
-		{ { "uk" }, "en", "eu" },
+		{ { "uk", "gb"}, "en", "eu"},
 		{ { "es", "spain", "s" }, "es", "eu" },
 		{ { "fr", "france", "fre", "french", "f" }, "fr", "eu" },
 		{ { "de", "germany", "d" }, "de", "eu" },
@@ -106,6 +101,14 @@ void LangInfo::extractLang(std::string val)
 	std::string systemLang = getSystemLang(locale);
 	std::string systemRegion = getSystemRegion(locale);
 
+	for (auto langData : langDatas)
+	{
+		if (std::find(langData.value.cbegin(), langData.value.cend(), systemRegion) != langData.value.cend())
+		{
+			systemRegion = langData.region;
+		}
+	}
+
 	for (auto s : Utils::String::splitAny(val, "_, "))
 	{
 		bool clearLang = s.find("t-") != std::string::npos;
@@ -134,16 +137,10 @@ void LangInfo::extractLang(std::string val)
 						region = langData.region;
 						mHardRegion = newHardRegion;
 					}
-					else if (newHardRegion)
+					else if (newHardRegion && region != systemRegion)
 					{
-						if ((region != systemRegion) && langData.region == systemRegion)
-						{
-							region = langData.region;
-						}
-						else if (((region != systemRegion) || systemRegion.empty()) && (langData.lang == systemLang))
-						{
-							region = langData.region;
-						}
+						if (langData.region == systemRegion || langData.lang == systemLang)
+							region = langData.region;						
 					}				
 				}
 			}
--- a/es-app/src/LangParser.cpp
+++ b/es-app/src/LangParser.cpp
@@ -30,23 +30,32 @@ std::string LangInfo::getLanguageString()
 	return data;
 }
 
+std::string getSystemLang(std::string locale)
+{
+	if (locale.empty())
+		return "en";
+	else
+	{
+		auto shortNameDivider = locale.find("_");
+		if (shortNameDivider != std::string::npos)
+			return Utils::String::toLower(locale.substr(0, shortNameDivider));
+	}
+}
+
 std::string getSystemRegion(std::string locale)
 {
-	if (locale == "en_US" || locale == "en")
-		return "us";
-	else if (locale == "en_GB")
-		return "eu";
-	else if (locale == "eu_ES")
-		return "eu";
-	else if (locale == "ja_JP")
-		return "jp";
-	else if (locale == "ko_KR" || locale == "ko")
-		return "kr";
-	else if (locale == "pt_BR")
-		return "br";
-	else if (locale == "pt_PT")
-		return "eu";
-	return std::string();
+	std::string region = std::string();
+	if (!locale.empty())
+	{
+		auto shortNameDivider = locale.find("_");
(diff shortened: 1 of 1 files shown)


geogolem:
> It's a bad solution to use getString("ThemeRegionName") in the extractLang method, most themes don't use/define regions. It would work with a very limited number of themes.

I totally get what you are saying. I was trying to make use of the existing options etc. so as to avoid having the user specify a preferred Region --> since this is sort of already specified in the GUI Options -- at least when I was testing -- but I am not that familiar with too many themes.. etc.

Having the check for the locale actually resolves all my test issues where I was running into this problem i.e. 1080 Snowboarding (Japan, USA) (En,Ja)

The locale check would make it pick USA over Japan for me or vice versa but I anticipated cases where there could be both USA and Europe in the region list.. which is possible in the nointro naming scheme.. Then the locale specification wouldn't be enough to differentiate.. whereas the themeregion settings which I saw in the GUI options on Batocera would...

in [es-core/src/ThemeData.cpp](https://github.com/batocera-linux/batocera-emulationstation/blob/27b3594d68ffb68f28cde23cb43c350e5023a561/es-core/src/ThemeData.cpp#L686)
there is a check and a default region used: 
```
mRegion = Settings::getInstance()->getString("ThemeRegionName");
if (mRegion.empty()) mRegion = "eu";
```
Maybe having a similar check to handle this case would improve the code. 

I need to have a look at how Batocera treats this ThemeRegion setting when using different themes --> if the setting is still under GUI Options.. then it should work fine --> but should still have the check for if its empty, then dont use it (i.e. falling back on the locale specification only)

I appreciate your feedback --> I'll look more into it later tonight or tomorrow. 

Just thought I would respond indicating my thoughts on it in case you had any further feedback or suggestions.. I mean, ultimately we could make a distinct option for region preference specification that could be used.. but I thought it would be better to piggy back on the existing options if possible... maybe.

fabricecaruso:
I don't think it's necessary to add a new option for the "prefered region". It's already possible to deduct the region based on "system.language". Creating a table of regions for this value should be enough ( en_GB : it's Europe... en or empty is US. pt_PT is Europe. pt_BR is brazil, eu_ES is europe, es_MX is south america... )

In my POV, Either :
- You make it simple and extract the user's region from the system.language value, then import the best region based on this data when you have multiple regions.
- Either you import every possible regions into a comma delimited string ( or ; ). In that case, it required some other changes in the code : the region flag image is to be computed : either it's the best calculated one (using ThemeRegion) or the one from system.language, either the first one.

geogolem:
> * You make it simple and extract the user's region from the system.language value, then import the best region based on this data when you have multiple regions.

> * Either you import every possible regions into a comma delimited string ( or ; ). In that case, it required some other changes in the code : the region flag image is to be computed : either it's the best calculated one (using ThemeRegion) or the one from system.language, either the first one.

I think storing all the possible regions into a comma delimited string is the better solution simply for completeness; however, just importing the best region using the locale info is still an improvement and more than "good enough" in my opinion...

fabricecaruso:
With French language (and many others), getSystemRegion returns an empty string, so your code is not working with many regions.

geogolem:
> With French language (and many others), getSystemRegion returns an empty string, so your code is not working with many regions.

its been a minute since I looked at this.. so I'll need to confirm what I'm writing here.. 

but getSystemRegion only needs to handle those regions where languages could correspond to multiple regions or where the region cannot be immediately determined from the language. 

With Korea the issue is Kr vs. Ko etc. and with Portugeses it can be either Brazil or European depending on pt_PT, or pt_BR

For example, with French, the langData specification specifies Europe as the region for language fr. 

this is taken care of in the second branch of the if statement:
```
						if ((region != systemRegion) && langData.region == systemRegion)
						{
							region = langData.region;
							}
						else if (((region != systemRegion) || systemRegion.empty()) && (langData.lang == systemLang))
						{
							region = langData.region;
						}
```

for example, with english(us) vs. english(gb) just the language is not enough to determine the region.. therefore getSystemRegion needs to handle that.. with fr_Fr, just the language fr is enough to determine the region in langData.

The idea is that getSystemRegion doesn't need to be exhuastive.. if it returns an empty string, it all still works.. because the systemLang is used instead.

For example, French should still all work as expected.. I think.. as i said its been a while so I will look more into it at some point.

geogolem:
> With French language (and many others), getSystemRegion returns an empty string, so your code is not working with many regions.

its been a minute since I did this --> however, the fact that it returns an empty string for many cases is intended... as the code fallse back on systemLanguage if systemRegion is empty... In most cases, the proper region can be deduced from the language and is already specified in langData --> however, in some cases, like english the preferred region cannot be determined.. i.e. is it eu or us?

in those cases, it is necessary... having said that.. I will commit a few changes that will make the code more complete although functionally won't change anything..

geogolem:
The code now basially will implement every region/language combination as specified in: 
[es-app/src/guis/GuiMenu.cpp](https://github.com/batocera-linux/batocera-emulationstation/blob/aa240ebac79c6a2463d6dd01ad4d4498fdafc858/es-app/src/guis/GuiMenu.cpp#L1124)

fabricecaruso:
Now, getSystemRegion returns "_fr", so it will probably never work for every language except for "gb", "pt" and "es"...
Also, getSystemLang has no default "return".
You also don't need to "else" a condition that performs a "return".

Testing your code with about 20000 filesnames, I have NO case where the code reaches one of your new assignements with most languages ( fr, it, de... ) and I have only one case with US english.

Can you provide a list of working test cases ?

geogolem:
> Now, getSystemRegion returns "_fr", so it will probably never work for every language except for "gb", "pt" and "es"... Also, getSystemLang has no default "return". You also don't need to "else" a condition that performs a "return".
> 
> Testing your code with about 20000 filesnames, I have NO case where the code reaches one of your new assignements with most languages ( fr, it, de... ) and I have only one case with US english.
> 
> Can you provide a list of working test cases ?

in the most recent commit be fore your comment I neglected to strip off the leading _ in the language which is the reason for the _fr which is why most of your tests hit the new assignments.

I since pushed 2 commits. First stipping off the leading _ and a subsequent one that better handles regions like Fr etc.

Below are examples of the types of cases, I personally aimed to fix: 
1080 Snowboarding (Japan, USA) (En,Ja)  
Without my code, region is Japan (first/only hardRegion detected).
With my code hardRegion USA is also detected:
In the case of English(US), USA is selected based on matching systemRegion (first branch of if statement).
In the case of English(UK), USA is selected based on matching systemLang (second branch of if statement)
For other locales like French, neither the region or language match so the behavior is unchanged from without my code. i.e. region is Japan which was the first one found.

Alex Kidd in Miracle World (USA, Europe, Brazil) (En) (Rev 1)
Without my code, region is USA (first/only hardRegion detected).
With my code hardRegions Europe and Brazil are also detected:
In the case of English(US), USA will be selected based on matching systemRegion
In the case of English(UK) Europe will be selected based on matching systemRegion
In the case of pt_BR, Brazil will be selected
For other locales like French, Europe will be selected based on matching systemRegion (note the systemRegion is adjusted based on langData.region) - since fr_Fr for the locale doesn't actually specify EU; however, for langData matching fr the langData.region is Eu.

The master system actually had many other examples similar to the last one.. but for the n64 the 1080 snowboarding is the only one I initially had an issue with...

geogolem:
I cleaned up the code a bit.. as I dont really see a need to have functions getSystemLang or getSystemRegion etc. i.e. basically it is just splitting the string at _ and the first part if the language, second part is the region and then the region needs to be adjusted based on the langData's used in the existing region detection code.

geogolem:
The code is alot simpler now and it now implements cases like fr corresponding to eu properly which I had previously overlooked.

geogolem:
The code is alot simpler now and it properly implements cases like fr corresponding to eu properly which I had previously overlooked.

I appreciate your feedback. It had been a minute since I originally did this, but your comments/feedback made me re-evaluate how it was implemented.. Hopefully it makes more sense now.

I am writing up a test suite.. so stay tuned for that...

EDIT: also --> the code handles the case where the system locale is for example English(UK) and the romname is romName (Japan, USA)... region US will be chosen not because the systemRegion matches directly but because no direct systemRegion match was found but USA corresponds to english which does match the system language..

small test suite coming soon --> just writing/verifying it now... will post again soon.

geogolem:
Test rom files:

romName1 (Japan, USA).ext
romName2 (Japan, USA, Europe, Brazil).ext
romName3 (Japan, Brazil, Europe).ext

Before the code changes, all of these would set region to Japan regardless of System Locale.

SystemLocale --> English(US) --> en_US
romName1 (Japan, USA).ext ---> region ==> USA
romName2 (Japan, USA, Europe, Brazil).ext ---> region ==> USA
romName3 (Japan, Brazil, Europe).ext ---> region ==> Europe

systemLocale --> English(UK) --> en_GB
romName1 (Japan, USA).ext ---> region ==> USA
romName2 (Japan, USA, Europe, Brazil).ext ---> region ==> Europe
romName3 (Japan, Brazil, Europe).ext ---> region ==> Europe

systemLocale --> FRANÇAIS--> fr_FR
romName1 (Japan, USA).ext ---> region ==> Japan
romName2 (Japan, USA, Europe, Brazil).ext ---> region ==> Europe
romName3 (Japan, Brazil, Europe).ext ---> region ==> Europe

systemLocale --> Japanese --> ja_JP
romName1 (Japan, USA).ext ---> region ==> Japan
romName2 (Japan, USA, Europe, Brazil).ext ---> region ==> Japan
romName3 (Japan, Brazil, Europe).ext ---> region ==> Japan

systemLocale --> PORTUGUÊS BRASILEIRO--> pt_BR
romName1 (Japan, USA).ext ---> region ==> Japan
romName2 (Japan, USA, Europe, Brazil).ext ---> region ==> Brazil
romName3 (Japan, Brazil, Europe).ext ---> region ==> Brazil

systemLocale --> PORTUGUÊS PORTUGAL--> pt_PT
romName1 (Japan, USA).ext ---> region ==> Japan
romName2 (Japan, USA, Europe, Brazil).ext ---> region ==> Europe
romName3 (Japan, Brazil, Europe).ext ---> region ==> Europe

Issue #1701, opened 2024-03-11
Feature Request


geogolem:
### Issue description

I'm not sure if this is the right place to open this issue.. but maybe you can provide some feedback and/or point me in the right direction..

First of all I am using Retrobat - but some of the discussion could also be applied to Batocera as well.

In Retrobat there is an option to autoconfigure the controllers which seems to be pretty useful BUT, its not easy to setup all my controllers this way in retrobat/emulation station so I prefer to turn that off and use retroarchs autoconfig. I can get all my controllers working no problem in retroarch using the most recent SDL2... 

The problem is it is not straightforward to map a 6 button Genesis controller or an n64 NSO controller in retrobat in such a way that it will then work properly in retroarch via retrobats autoconfig option.. so I disabled the autoconfig option and have all my controllers worker great inside retroarch.. I suppose if I could properly map everything in Retrobat I could re-enable the autoconfig and it would work fine too..but right now I cant.. (maybe the latest version of SDL2.dll would also resolve some of the issues I was having, but I am not sure)...

Anyway, even it might also be possible to map everything by manually editing the es_input.cfg file as well.. but for now I have disabled autoconfig in retrobat.

Having said all that.. my feature request is more related to the UI buttons.. I know there are options to show A/B or based on position.. I use the "based on position" buttons but the feature request is regarding the ability of retrobat to detect which controller is being used.. i.e. when navigating retrobat why don't the UI elements for the buttons change depending on the active controller... i.e. retrobat definitely can tell the difference between all my controllers... so it would be awesome if it could change the image being shown in realtime depending on the active controller.

geogolem:
This isn't complete but it is pretty awesome: https://opengameart.org/content/free-keyboard-and-controllers-prompts-pack
https://www.reddit.com/r/gamedev/comments/1z0zid/free_keyboard_and_controllers_prompts_pack/

geogolem:
maybe it could also be some kind of function.. i.e. tap hotkey on a controller, boom, the GUI changes to show the help icons corresponding to that controller..  pair another controller.. tap hotkey on it.. boom the GUI changes to show the help icons corresponding to that controller.. 

[...]

PR #1706, opened 2024-03-15
Feature: Allow theme to Dynamically change HelpIcons to match active controller

geogolem:
see discussion here: https://github.com/batocera-linux/batocera-emulationstation/issues/1701

I realize the implementation using the externs for ActiveController and ActiveControllerGUID are probably not perferred.. but I just wanted to open the PR to get some feedback on what I've done so far. It seems to work well for me and doesn't impact the existing implementation/code if the feature is disabled (or even enabled but not implemented by the theme)..

Any feedback is greatly apprecaited!

Thanks in advance!

### DIFF
--- a/es-core/src/HelpStyle.cpp
+++ b/es-core/src/HelpStyle.cpp
@@ -1,6 +1,11 @@
 #include "HelpStyle.h"
 
 #include "resources/Font.h"
+#include "Settings.h"
+
+extern std::string activeControllerName;
+extern int activeControllerVID;
+extern int activeControllerPID;
 
 HelpStyle::HelpStyle()
 {
@@ -15,6 +20,49 @@ HelpStyle::HelpStyle()
 
 	if (FONT_SIZE_SMALL != 0)
 		font = Font::get(FONT_SIZE_SMALL);
+}
+
+auto HelpStyle::applyDynamicHelpSystem(const std::shared_ptr<ThemeData>& theme, const std::string& view)
+{
+
+		std::string activeControllerVIDString = "";
+
+		switch(activeControllerVID)
+		{
+			case 1406:
+				activeControllerVIDString = "Nintendo";
+				break;
+			case 1356:
+				activeControllerVIDString = "Sony";
+				break;
+			case 1118:
+				activeControllerVIDString = "Microsoft";
+				break;
+		}
+
+		//try Vendor/Name, example, Nintendo/SNES_Controller
+		auto elem = theme->getElement(view, activeControllerVIDString+"/"+activeControllerName, "helpsystem");
+		if (elem)
+			return elem;
+		
+		//try VID/Name, example, 1406/SNES_Controller
+		activeControllerVIDString = std::to_string(activeControllerVID);
+		elem = theme->getElement(view, activeControllerVIDString+"/"+activeControllerName, "helpsystem");
+		if (elem)
+			return elem;
+
+		std::string activeControllerPIDString = "";
+		activeControllerPIDString = std::to_string(activeControllerPID);
+		
+		//try VID/PID, example, 1406/8215
+		elem = theme->getElement(view, activeControllerVIDString+"/"+activeControllerPIDString, "helpsystem");
+		if (elem)
+			return elem;
+
+		//try deviceName only, example, SNES_Controller
+		elem = theme->getElement(view, activeControllerName, "helpsystem");
+
+		return elem;
 }
 
 void HelpStyle::applyTheme(const std::shared_ptr<ThemeData>& theme, const std::string& view)
@@ -47,6 +95,13 @@ void HelpStyle::applyTheme(const std::shared_ptr<ThemeData>& theme, const std::s
 
 		if (elem->has("fontPath") || elem->has("fontSize"))
 			font = Font::getFromTheme(elem, ThemeFlags::ALL, font);
+
+		if (Settings::getInstance()->getString("subset.dynamichelpsystem") == "yes")
+		{
+			auto elem2 = applyDynamicHelpSystem(theme, view);
+			if (elem2)
+				elem = elem2;
+		}
 
 		if (elem->has("iconUpDown"))
 			iconMap["up/down"] = elem->get<std::string>("iconUpDown");
--- a/es-core/src/HelpStyle.cpp
+++ b/es-core/src/HelpStyle.cpp
@@ -68,6 +68,13 @@ auto HelpStyle::applyDynamicHelpSystem(const std::shared_ptr<ThemeData>& theme, 
 void HelpStyle::applyTheme(const std::shared_ptr<ThemeData>& theme, const std::string& view)
 {
 	auto elem = theme->getElement(view, "help", "helpsystem");
+	if (Settings::getInstance()->getString("subset.dynamichelpsystem") == "yes")
+	{
+		auto elem2 = applyDynamicHelpSystem(theme, view);
+		if (elem2)
+			elem = elem2;
+	}
+
 	if (elem)
 	{
 		if (elem->has("pos"))
@@ -95,13 +102,6 @@ void HelpStyle::applyTheme(const std::shared_ptr<ThemeData>& theme, const std::s
 
 		if (elem->has("fontPath") || elem->has("fontSize"))
 			font = Font::getFromTheme(elem, ThemeFlags::ALL, font);
-
-		if (Settings::getInstance()->getString("subset.dynamichelpsystem") == "yes")
-		{
-			auto elem2 = applyDynamicHelpSystem(theme, view);
-			if (elem2)
-				elem = elem2;
-		}
 
 		if (elem->has("iconUpDown"))
 			iconMap["up/down"] = elem->get<std::string>("iconUpDown");
--- a/es-core/src/HelpStyle.h
+++ b/es-core/src/HelpStyle.h
@@ -26,6 +26,7 @@ struct HelpStyle
 	
 	HelpStyle(); // default values
 	void applyTheme(const std::shared_ptr<ThemeData>& theme, const std::string& view);
+	auto applyDynamicHelpSystem(const std::shared_ptr<ThemeData>& theme, const std::string& view);
 };
 
 #endif // ES_CORE_HELP_STYLE_H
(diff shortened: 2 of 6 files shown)


fabricecaruso:
I did not anwered your "issue" because I see more complications than interests, and I don't want to spend hours discussing this.

About the "concept" :

- There is not good way to detect what icons best suits a controller. I could make sense ony with some controllers like xbox or playstation controllers : Not others. 
- Even with a Xbox/PS (compatible) controller ( like a 8bitdo ), you can configure it with xbox controller conventions, but you can also also configure it using snes or PS convention : users do as they want and you have no good way to detect such customizations.
- HelpSystem icons are currently managed by themes and themes provide (or not) the icon suite they want. Most theme don't have that kind of choices, there's often only one choice. 
- Themes can also implement variants for the same controller type. If you detect the controller and automatically select a "helpsystem", you can't choose variants anymore in the theme options.
- If you add a new global option, but no theme implement it, you just have created a useless option. You have something in the menus that nobody will ever understand. So you should not have the option if the theme does not support the feature.

For those reasons, I don't think this is not good idea.

Now, about the submitted implementation :

Honestly ? a "Name + Guid" based themed <helpsystem> name !?!?

Do you really know what's behind a SDL guid ? 
In a basic gamescontrollerdb file you have more than 2000 differents guids...
Do you really think someone will create a maintain a theme having more than 2000 <helpsystem> entries ?

I could make ( a little bit ) sense if that was : 
- Limited to only 3 cases : xbox / playstation / nes using pid/vid (which will be hard to manage for compatible controllers).
- If the "name" convention was basic (xbox / playstation / nes ? ), which will be simple to document and implement later for themers.
- If the menu option was hidden if the theme does not support the feature.

geogolem:
Thank-you for taking the time to respond!! It is greatly appreciated.

I totally get what you are saying and actually agree on many points - but I still personally do really like this feature (but I understand if it never gets into official builds -- but am also interested in following up/improving until it is in a form that could be accepted if possible etc.)

Below are some of my initial thoughts regarding your comments but I really need to think things over to see if I can come up with a more elegant, still flexible but simpler solution. 

> To finish : I could make sense if that was :
> 
> * Limited to only 3 cases : xbox / playstation / nes using pid/vid for detection (which will be hard to manage for compatible controllers).

We could use PID/VID but there are "incompatible" devices that sometimes share PID/VID.. For example, the NSO SNES, NSO 3 button Genesis and NSO 6 button Megadrive controller all share the same PID/VID... There are other cases as well.

My use case is basically the following controllers:

Official PS4 or PS5 controller
NSO SNES controller
NSO NES controller
NSO 6button Megadrive Controller
NSO N64 Controller
latest XBOX controller (i dont own this one yet)

I also wouldn't mind having help icons for the keyboard which would be easy to implement via the theme but for that one we could just use name "keyboard" etc. and not GUID etc.. this one is actually simpler..

> * If the "name" convention was a short list of values like "xbox",  "playstation", and "nes", which is simple to document and implement for themers.

I get the desire to keep it simple for themers --> but I also place value on the ability to handle any specific controller without having to modify the code etc. If the name convention is made to be too simple, then the actual controller handling needs to happen in the code --> then if someone has a controller they really want to have specific helpIcons for, they could simply handle it by adding an entry to dynamicbuttons.xml and supplying the corresponding image files.

I think a combination of both methods is probably possible... one that basically can allow the theme to create/specify simple names to use.. I mean it may be possible to totally just forget about GUID and just use the deviceName which is simpler.. I know this will work for my controllers... just deviceName... 
something similar to: https://www.retroarch.com/index.php?page=controller-autoconfig

> * If the menu option was hidden if the theme does not support the feature.

I did just force push a change that basically moves the option and everything related to it into the theme itself.. so basically if a theme doesn't provide this option, it doesn't appear etc. as you suggested.. I agree with this idea.

Anyway I need to think about this some more but I won't write back anymore unless I feel that I have adequately addressed all your points...

geogolem:
Thank-you for taking the time to respond! It is greatly appreciated.

I understand all your points, especially the GUID issue.. I have implemented some changes in response to your feedback.

The entire option is now only available/exposed if a theme implements it.

GUID is no longer used... but rather some combination of VID/PID/deviceName

Referring to:  https://www.retroarch.com/index.php?page=controller-autoconfig
```
RetroArch provides a set of configuration files for the most common controllers. Depending on your platform, these autoconfig profiles are either distributed with RetroArch or downloaded through the Online Updater.RetroArch provides a set of configuration files for the most common controllers. Depending on your platform, these autoconfig profiles are either distributed with RetroArch or downloaded through the Online Updater. Matching is done using 3 criteria: the device name, the vendor id and the product id. The vendor id and product id pair is often abbreviated as vid:pid. We compute a matching score for each configuration file based on these three factors. The autoconfig profile with the highest score is used to map the controller.
```
and looking in the code for retroach (referring to: https://github.com/libretro/RetroArch/blob/cc97df732cc60271ca0cac996950101ee56eb376/tasks/task_autodetect.c#L118):
```
/* Returns a value corresponding to the
 * 'affinity' between the connected input
 * device and the specified config file
 * > 0: No match
 * > 2: Device name matches
 * > 3: VID+PID match
 * > 5: Both device name and VID+PID match */
 ```
I thought it best to implement it similar to that so I implemented it like this:
```
auto HelpStyle::applyDynamicHelpSystem(const std::shared_ptr<ThemeData>& theme, const std::string& view)
{

		std::string activeControllerVIDString = "";

		switch(activeControllerVID)
		{
			case 1406:
				activeControllerVIDString = "Nintendo";
				break;
			case 1356:
				activeControllerVIDString = "Sony";
				break;
			case 1118:
				activeControllerVIDString = "Microsoft";
				break;
		}

		//try Vendor/Name, example, Nintendo/SNES_Controller
		auto elem = theme->getElement(view, activeControllerVIDString+"/"+activeControllerName, "helpsystem");
		if (elem)
			return elem;

		//try VID/Name, example, 1406/SNES_Controller
		activeControllerVIDString = std::to_string(activeControllerVID);
		elem = theme->getElement(view, activeControllerVIDString+"/"+activeControllerName, "helpsystem");
		if (elem)
			return elem;

		std::string activeControllerPIDString = "";
		activeControllerPIDString = std::to_string(activeControllerPID);

		//try VID/PID, example, 1406/8215
		elem = theme->getElement(view, activeControllerVIDString+"/"+activeControllerPIDString, "helpsystem");
		if (elem)
			return elem;

		//try deviceName only, example, SNES_Controller
		elem = theme->getElement(view, activeControllerName, "helpsystem");

		return elem;
```

I do think this mostly addresses your 3 points but still leaves it open ended enough that if a user or themer wanted to implement for a specific controller they still could, but they could also implement simply for controllers of interest.. For example all official controllers from the big 3 etc. For example, I would probably also implement keyboard helpsystem in the theme as well which this allows.

Anyway, I am still thinking about ways to improve/simplify it.. but I also want to make it powerful/flexible enough that it could in theory handle even odd cases, if a themer/user wanted to do something for a unique/special controller.

Again -- as always your feedback is greatly appreciated.

fabricecaruso:
Well, you should refactor (many) things.

- I don't like the InputManager::getDeviceVID thing at All. It's a mistake to expose those method, in a class that is not designed for that, and ask the index to external code.
Instead, you should handle the controller type (xbox/ps/nes) as an enum ( or eventually a string ) in the InputConfig class. 
That would be reusable and the location make sense.
-> create some method like ControllerType InputConfig::getControllerType() that return the controller type.

- Don't call SDL methods at every frame ( SDL_JoystickGetDeviceVendor ).
Either you call it only one time and cache it, in the InputConfig class. Either :
It's not necessary to call SDL_JoystickGetDeviceVendor / SDL_JoystickGetDeviceProduct to extract the pid/vid : 
You can to it extracting this from the guid ( if you have 030000005E040000E002000000007200, 5e04 is the vendor, E002 is the vendor - revert the string, extract required chars, convert from hex to decimal and you're good )

- The activeControllerName/activeControllerVID global variables in Window class ( and in HelpStyle as extern ) are bad. 
You must get rid of them, once you have finished writing InputConfig::getControllerType().

-  Where does "subset.dynamichelpsystem" comes from ? There's no menu option anymore. It's not related to the theme either..

- You really should simplify applyDynamicHelpSystem() method -> there's no need to test "Nintendo/SNES_Controller", then "1406/SNES_Controller", then "1406/8215", then "SNES_Controller". Simply test the controller type you have from getControllerType() ( xbox/ps/nes ). That's enough. 

Think about themers : how will they guess what's the good name to use if it's complex as you did ? You need to keep it very simple. !

<helpsystem name="nes"> or <helpsystem name="xbox"> should be more than enough.

geogolem:
> * I don't like the InputManager::getDeviceVID thing at All. It's a mistake to expose those method, in a class that is not designed for that, and ask the index to external code.
>   Instead, you should handle the controller type (xbox/ps/nes) as an enum ( or eventually a string ) in the InputConfig class.
>   That would be reusable and the location make sense.
>   -> create some method like ControllerType InputConfig::getControllerType() that return the controller type.
> * Don't call SDL methods at every frame ( SDL_JoystickGetDeviceVendor ).
>   Either you call it only one time and cache it, in the InputConfig class. Either :
>   It's not necessary to call SDL_JoystickGetDeviceVendor / SDL_JoystickGetDeviceProduct to extract the pid/vid :
>   You can to it extracting this from the guid ( if you have 030000005E040000E002000000007200, 5e04 is the vendor, E002 is the vendor - revert the string, extract required chars, convert from hex to decimal and you're good )

I fixed that in the last commit.. it now basically just strips the bytes of the GUID.


> * The activeControllerName/activeControllerVID global variables in Window class ( and in HelpStyle as extern ) are bad.
>   You must get rid of them, once you have finished writing InputConfig::getControllerType().
I realize this.. and it still needs work -- still need to think about this and the other points you made...

geogolem:
Just wanted to note this here as it seems like an interesting project relevant to this feature: https://github.com/rsubtil/controller_icons

geogolem:
Just wanted to link to this which seems very interesting/relevant; however it uses SDL3.. https://github.com/icculus/ControllerImage

geogolem:
I decided to close this PR as a result of that SDL3 library that would make implementing this feature more feasible. The issue is still open and further discussion can be made here; https://github.com/batocera-linux/batocera-emulationstation/issues/1701

PR #1718, opened 2024-04-18
Steering wheel optimization (part 5)

Tovarichtch:
Optimizing arcades rotation degrees. Not perfect science, mainly based on arcade manuals, real-world steering knowledge (potentiometer, paddle, etc) and testing.

Infinite turn wheel games like "Championship Sprint" are set to 360 to keep it playable for high end modern steering wheel. Remove it for DD (direct drive) wheels (e.g.: MOZA Racing R9) to get full accuracy.

### DIFF
--- a/resources/gamesdb.xml
+++ b/resources/gamesdb.xml
@@ -23,7 +23,7 @@
       <gun/>
     </game>
     <game name="500gp">
-      <wheel/>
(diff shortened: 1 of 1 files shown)


Tovarichtch:
OK to merge. Removed few non-wheel games.

PR #1732, opened 2024-05-11
Add Daphne non-hd/Singe 2 gun games compatibility

Tovarichtch:
Non-HD and Singe2 ROMs from the official Hypseus Singe Data repository are not present in the game database since its merged. This will add their compatibility to be properly detected as gun games.

### DIFF
--- a/resources/gamesdb.xml
+++ b/resources/gamesdb.xml
@@ -366,6 +366,12 @@
       <gun/>
       <wheel rotation="270"/>
     </game>
+    <game name="CP_HD">
+      <gun/>
+    </game>
+    <game name="CP2DW_HD">
+      <gun/>
+    </game>
     <game name="cracksht">
       <gun/>
     </game>
@@ -376,6 +382,9 @@
       <gun/>
     </game>
     <game name="crimepat">
+      <gun/>
+    </game>
+    <game name="crimepatrol">
       <gun/>
     </game>
     <game name="crimepatrol-hd">
@@ -653,6 +662,9 @@
     </game>
     <game name="driveyes">
       <wheel rotation="270"/>
+    </game>
+    <game name="drugwars">
+      <gun/>
     </game>
     <game name="drugwars-hd">
       <gun/>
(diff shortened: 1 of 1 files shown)


nadenislamarre:
"CP_HD" => i think it will not be found. no _ and lowercase.

Tovarichtch:
Sadly, I've tried lowercase and without underscore `_` but it won't detect the folder as a gun game. Only uppercase with underscore works. Here is the structure used by [Hypseus Singe data repository](https://github.com/DirtBagXon/hypseus_singe_data/tree/master/00-singe2) for Singe2 games:

image

Maybe is it because the ROM is a folder instead of a single file, therefore it must match the exact same name with underscore and capital letters? This issue needs to be addressed, otherwise users will require to rename their folder game name *and* modify the scripts from Hypseus Singe data repository so that Batocera can properly launch the games. This is not practical and we need standardization.

Tovarichtch:
I would like to add that I think removing any other names would force users to stick with official assets from the repository.

E.g.: `spacepirates-hd` is not an official ROM name at all. I've added it because I was using it personally, but this is adding confusion to ROM names. Sticking with Singe and Singe 2 official folder names should be the norm.

Tovarichtch:
@fabricecaruso do you have any idea for the best way to add the names from this PR?

In short and as an example, here's the full list of Mad Dog McCree ROM names:

`maddog` `maddoga` `maddog_203` are all MAME ROM names, currently non-emulated.
`maddog` is SINGE1 ROM name.
`maddog-hd` is SINGE1 to SINGE2 port ROM name with minimal overlay.
`Maddog_HD` is SINGE2 ROM name with new assets and HD overlay.

As of today, only `Maddog_HD` is missing from the game database. If someone is using the newest assets from SINGE (available [here](https://github.com/DirtBagXon/hypseus_singe_data)), there'll be no gun icon because of the missing ROM name. Same case for Crime Patrol 2: Drug Wars with `CP2DW_HD` and all other SINGE2 gun games. Using lowercap and no underscore doesn't flag the games with the gun icon. I've tried `maddoghd` and `cp2dwhd` to no avail.

What do you suggest?

PR #1738, opened 2024-06-01
fix switch on / off.

dmanlfc:
- be clearer on the condition
- also for switchoff, write config to 1 only when toggled on (true).


nadenislamarre:
for "be clearer on the condition", personnally, i prefered the previous write, but if this write helps, it is ok to me.

PR #1743, opened 2024-06-13
Add teknoparrot and xbox gun games

Tartifless:


resources/gamesdb.xml
```diff
@@ -6551,6 +6673,15 @@
     <game name="ihraprofessionaldragracing">
       <wheel wheel="joystick1left" accelerate="lt" brake="rt"/>
     </game>
+    <game name="silentscopecomplete">
+      <gun/>
+    </game>
+    <game name="starskyandhutch">
+      <gun/>
+    </game>
+    <game name="thehouseofthedeadiii">
```
Tovarichtch, line 6682: Remove `the` at the beginning. This is a known issue with the scraper. The game will be understood as `House of the Dead III, The`, not the otherway around.

Tovarichtch (changes requested):

resources/gamesdb.xml
Tartifless: Done

PR #1744, opened 2024-06-17
fix battery < 10%

dmanlfc:


fabricecaruso:
I has to revert this commit as it's not compiling well on windows, and as I don't understand how it fixes anything

dmanlfc:
currently `Utils::FileSystem::readAllText` is not reading the values correctly.
hence adding a new method which reads the integer from the file correctly.
this needs to be resolved in one way or another. ping me on Discord.

PR #1760, opened 2024-07-18
Fix teknoparrot titles and add gun + wheel games

Tartifless:
Uppercase is necessary, else ES does not find the games

### DIFF
--- a/resources/gamesdb.xml
+++ b/resources/gamesdb.xml
@@ -19,6 +19,9 @@
     <game name="2spicy">
       <gun/>
     </game>
+    <game name="2Spicy">
+      <gun/>
+    </game>
     <game name="38ambushalley">
       <gun/>
     </game>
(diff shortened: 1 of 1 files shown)


Tovarichtch:
Similar issue with #1732 about uppercases. As long as it doesn't touch generated `arcaderoms.xml` file, it's fine.

PR looks good. However, this is optional but futureproof: do you know the max wheel rotation for each of these systems? Or TeknoParrot doesn't care?

resources/gamesdb.xml
```diff
@@ -19,6 +19,9 @@
     <game name="2spicy">
       <gun/>
     </game>
+    <game name="2Spicy">
+      <gun/>
+    </game>
```
Tovarichtch, line 24: Duplicate?

Tartifless: No, uppercase

Tartifless:
TP does not care about max wheel rotation, neither does windows generally

resources/gamesdb.xml
Tovarichtch: I see two times the game name. One uppercase, one lowercase. Game is both names?

Tartifless:
2spicy lowercase already existed, for mame

Tovarichtch:
All good then. OK to merge.

PR #1778, opened 2024-09-08
Slider : add a AUTO value completely on the left

Tartifless:
The auto value is completely on the left, it is also the default value when nothing is set and it saves nothing in conf file when on AUTO


nadenislamarre:
hum, -1.0 cannot be the default value otherwise, we can't set it.

Tartifless:
> hum, -1.0 cannot be the default value otherwise, we can't set it.

Hi, i have updated the code to get the AUTO value automatically based on preset-parameters (with getAuto () and setAuto() methods).

Tartifless:
@fabricecaruso : can you have a look if this is ok with you ?

fabricecaruso:
if we take the volume slider, which can't be "auto", well it has an auto position. It should not.
I also don't like it the display where the circle is still visible on the left side.

In my opinion it is necessary that:
- you make modifications so that the auto mode does not apply to existing sliders outside of presets like the volume. As long as the standard sliders do not activate an auto mode, it must not exist / be possible (cf. volumes, durations of notifications....)
- Maybe hide the circle when auto is active

PR #1792, opened 2024-09-15
Add sliderauto component

Tartifless:
Fixed :
- no impact on normal slider
- knob disappears when in AUTO


Tartifless:
@fabricecaruso 
I have created a new PR, cancelled the old one

PR #1793, opened 2024-09-18
Sliderauto: add calculation of precision based on slider_step + new preset for filebrowser

Tartifless:


Tartifless:
@fabricecaruso : i have updated the method to get the decimal places as discussed, this one is correct, i have tested with multiple decimals values and it gives me the correct figure without issues in floating point precision.

PR #1799, opened 2024-10-10
Improved Network Play Sequence

bulzipke:
z

RetroArch and EmulationStation have many convenient features already, but Netplay is still challenging for new users.

1. Hosting a Game:
- Previously, to host a Netplay game, users had to follow the steps shown in **screenshot 1**:
  - Select a game
  - **Hold the A button**
  - Choose HOST A NETPLAY GAME.
- The requirement to **hold the A button** for hosting is not intuitive for new users who may not have read the manual.

2. Joining an Existing Game:
- To join an existing game, as shown in **screenshot 2**, users press the X button from the main screen. This process has not been modified in my PR.

3. My PR Combines Hosting and Joining Options:
- In my PR, I have combined both functions (hosting and joining) into the **screenshot 3** feature.
- Now, when a user selects a game, they can immediately choose whether to:
  - PLAY ONLINE AS HOST,
  - PLAY OFFLINE, or
  - JOIN EXISTING GAME.
- The existing rooms displayed are only for the selected game, allowing users to focus on the game they intend to play.

**Code Modifications:**
The following changes were made to achieve this:
- When the ENABLE NETPLAY option is active, the screens for creating a room, joining a room, or offline play now appear when launching a game.
- The option that appears when holding down the A button during game launch has been moved to NETPLAY SETTINGS.
- The check for isNetplaySupported() has been removed from FileData::getlaunchCommand because rooms can now be created in WATCH mode (SPECTATOR) regardless of the core's support for netplay.

Please feel free to let me know if you need any further changes or have any questions.


Darknior:
I admit that it's not bad, faster and easier to use.
It only lists the game you want to play with your friend...

you just have to leave the screen of all the parts online because it is another use. But otherwise if you know what you're going to play with your friend, I admit it's much better :)

bulzipke:
> I admit that it's not bad, faster and easier to use. It only lists the game you want to play with your friend...
> 
> you just have to leave the screen of all the parts online because it is another use. But otherwise if you know what you're going to play with your friend, I admit it's much better :)

The screen showing all online rooms is still accessible through the X button on the main screen. Thanks for the feedback!

fabricecaruso:
Honnestly ???? I HATE this !!!!!!!!!!!!!!!!!!!!!!!! And I personnally don't think it is an improvment...

People almost never use netplay... 

And now, you always have a screen when you want to launch a game... Even if you play offline 99% of the time

Also, it's a total non-sense : The 1st option is to host a netplay game, when it's the less frequent option... Crazy !

Do somebody really think this PR is good ????????

I wouldn't have said anything if you added an option in the NETPLAY OPTIONS, to let user activate this screen or not ( with OFF by defaut ). ON -> You have that screen. OFF -> Works like before.

As there is no solutions to avoid this screen, in my optinion, this PR should be reverted and cancelled.

bulzipke:
> Honnestly ???? I HATE this !!!!!!!!!!!!!!!!!!!!!!!! And I personnally don't think it is an improvment...
> 
> People almost never use netplay...
> 
> And now, you always have a screen when you want to launch a game... Even if you play offline 99% of the time
> 
> Also, it's a total non-sense : The 1st option is to host a netplay game, when it's the less frequent option... Crazy !
> 
> Do somebody really think this PR is good ????????
> 
> I wouldn't have said anything if you added an option in the NETPLAY OPTIONS, to let user activate this screen or not ( with OFF by defaut ). ON -> You have that screen. OFF -> Works like before.
> 
> As there is no solutions to avoid this screen, in my optinion, this PR should be reverted and cancelled.

If you disable the NETPLAY option, you will not get the netplay screen as before.

Screenshot_20241029_150830_Chrome

fabricecaruso:
Yeah... I noticed. And it's a bad way to disable that screen.
You probably didn't understand why I dont like this PR.

With your code : 
- Either NETPLAY is ON and you always have that new screen everytime you want to launch a game. ( **Even if you want to launch an offline game from a savestate !!!!** )
- Either NETPLAY is OFF, you don't have the screen, but you can't host/join a NETPLAY game anymore, unless you go and check the option again, then have the screen always....

This is bad to me.
Most of the time, users have netplay configurated, but they play locally 99% of the time, and they don't want to have an intermediate screen, just because you want to be able to join a netplay game.

Let's put ourselves in the players' shoes : Users will use netplay probably less than 1% of the time.
Showing this new screen 100% of the time for a "use case" that is less than 1% is a very bad idea.
Once configurated, you have to "click" 3 times instead of 1 to launch a local game.

Also : force users to totally disable NETPLAY make it impossible to join a netplay game unless you activate the option again. And if you activate the option, you always have this screen !!!!

Imagine you want to join a game but you don't have NETPLAY configurated. You now have to make at least 12 actions to join the game... Instead of (at least) 2....

I still really don't like this and I'm not the only one  : I reacted here because I was told about this PR by the members of the Retrobat Team. They ALL told me that they did not like it at all, and I clearly agree with them.
This is clearly a bad idea when the use case is less than 1%

In my opinion : either you find a way (with an option ?) to have the NetPlay host sequence like it was before ( by default ), either this PR should be reverted. Also, the screen should probably be disabled when running from a savestate.

@Tartifless @Darknior @lbrpdx @nadenislamarre @dmanlfc @lorenzolamas 
Can you please give you point of view / vote ( thumb up / thumb down ) after having it tested ?

bulzipke:
Another RetroBat Team member had the idea, he suggested adding more options, what do you think?

https://github.com/batocera-linux/batocera-emulationstation/commit/4d7dfa826a4440d071feaf2f3340635f6a0b8303#commitcomment-148466810

bulzipke:
I have PR'd the addition of the option.
A review would be appreciated.

fabricecaruso:
I just fixed the newly introduced indexation problem. 
About the new option, it's fine to me in the current state

Darknior:
I think it’s a good idea to simplify the procedure.

Being able to access a screen with one click that allows you to host or list ongoing games for connection is really convenient.
BUT!!! As Fabrice said, it should be triggered by a long press or a trick like that... because even though I activate Netplay and use it sometimes, it's rare—about 5% of my games... so seeing this screen every time will be tedious.
And I don’t want to have to disable Netplay... so we can’t have this window popping up every time.
It needs to be done through a manipulation; a short press should just launch the game normally, nothing more.

bulzipke:
> I just fixed the newly introduced indexation problem. About the new option, it's fine to me in the current state

I don't think anyone is interested in this feature after the new release is up. 
I also think this would be a way to fragment the existing access to netplay, which could be confusing for users.

I'm really sorry to bother you, but would it be okay if I PR'd revert this feature along with the related commits?

bulzipke:
If that's okay, I'll request a revert for this PR, it's causing confusion for users.

PR #1813, opened 2024-10-27
Update emulationstation2.pot

DDinghoya:


nadenislamarre:
hi,
the pot file is an automatically updated file.
for batocera translations, go here : https://translations.batocera.org/

DDinghoya:
> hi, the pot file is an automatically updated file. for batocera translations, go here : https://translations.batocera.org/

Thank you for your strong support!

Previously I had committed to github twice and it was reflected. After being informed, I am registering according to the translation site's procedures.

I think my ID (DDinghoya) should be registered as a translator on https://translations.batocera.org/

Can you decide? Or is there another way?

nadenislamarre:
yes, for which language ?

DDinghoya:
> yes, for which language ?

Korean on translation site. :)

PR #1820, opened 2024-11-08
Clear text input field for passwords

n2qz:


n2qz:
After discussion today with @lbrpdx I've agreed to adjust this to only clear the password when the ENFORCE SECURITY option is enabled.

BloodyIron:
Thanks!

PR #1846, opened 2024-12-25
Generate or refresh cheevos token at startup

n2qz:
Batocera needs this to support fixes for #13111


fabricecaruso:
You should not do that in the main function, but in a thread instead !

PR #1847, opened 2024-12-26
Move cheevos token refresh to a thread

n2qz:
Addresses the feedback in https://github.com/batocera-linux/batocera-emulationstation/pull/1846#issuecomment-2562870242


fabricecaruso:
I moved your code to the NetWorkThread/WatchersManager.
Doing this, it's standardized, and the token is checked every 2 hours.

https://github.com/batocera-linux/batocera-emulationstation/pull/1848

PR #1854, opened 2025-01-15
Add 3 teknoparrot games (gun)

Tartifless:


Tartifless:
@Tovarichtch : ok for you ?

Tovarichtch:
Looks good. Profile Game ID of TP correct?

Tartifless:
Yes already tested in RetroBat by replacing the gamesdb file

dmanlfc:
any reason to have uppercase in the names?

Tovarichtch:
> any reason to have uppercase in the names?

Arcade database with wheel/gun is case sensitive and must be written as such. Otherwise, ES won't detect the game. Fabrice confirmed to susan34 and me that it was OK as long as it is only this file, and not the arcade one which is generated with MAME database.

Tartifless:
Case must be respected, Teknoparrot games are case-sensitive

Tartifless:
Can you merge these ? We are preparing release of v7 this week.

PR #1858, opened 2025-01-25
[PopupKeyboard] Add Korean & 4th Layout & fix Korean render bug & Unicode support for "Jump to game beginning with the letter"

bulzipke:
image (1)

Korean(Hangul) input is now supported, and the altedShifted state has been added, allowing for more diverse input.


bulzipke:
image
Fixed an issue where the DroidSansFallbackFull font had a bug rendering Korean characters, updated to handle it using fallback fonts.

UzuCore:
Recommend it.

bulzipke:
image
Fix incorrect character display for multi-byte characters in "Jump to game beginning with the letter" and implement Korean character support.

PR #1860, opened 2025-01-26
Add toggle to enable/disable Rumble

spycat88:
This currently only works with RP5/RPmini

### DIFF
--- a/es-app/src/ApiSystem.cpp
+++ b/es-app/src/ApiSystem.cpp
@@ -1363,6 +1363,54 @@ void ApiSystem::setBrightness(int value)
 	Utils::FileSystem::writeAllText(BACKLIGHT_BRIGHTNESS_NAME, content);
 }
 
+static std::string RUMBLE_NAME;
+
+bool ApiSystem::getRumble()
+{
+	#if WIN32
+	return false;
+	#endif
+
+	if (RUMBLE_NAME == "notfound")
+		return false;
+
+	if (RUMBLE_NAME.empty()) {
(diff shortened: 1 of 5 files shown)


nadenislamarre:
rumble of what ? the board itself ?

spycat88:
disables it in the driver directly for those who may want a silent gaming session

nadenislamarre:
is /sys/devices/platform/rumble_enable something standard ?
i don't find it on the linux documentation.
i fear it work only for one pad.

spycat88:
No its not however that is exactly why I set it up to search for rumble_enable as the path varies between drivers.

It actually works for all Allwinner, Rockchip and Qualcomm devices that support rumble in their joypad drivers.

Its not a standard Linux API path though but a lot of what gets done also isn't and a lot of features that various distros add are kept in downstream patches for various reasons.

This PR is a mere concept with request for comment, I'm happy to rework it in a way that it works best for everyone or drop it entirely and keep it confined to ROCKNIX, however we are trying to encourage everyone to upstream here so everyone is working from the same source and to keep forked changes as minimal as possible.

spycat88:
Worth adding that I kept the Allwinner and Rockchip paths out of this PR as the joypad driver that batocera is using for these devices does not support rumble however the RP5/mini one does as I already PR'd this previously.

Plan is to eventually rewrite the joypad drivers and make a universal one that works and plays nice with all handheld devices, I already have a PoC... queue OpenJoyPad.

spycat88:
Closing due to inactivity.

PR #1861, opened 2025-01-28
[Netplay] Reverted #1799 causing fragmentation

bulzipke:
It seems that no one is interested in this feature even in the new release, and it rather causes fragmented approaches to Netplay access, leading to confusion.

I have reverted #1799 and modified it to enable only the host option with the existing `netplay_public_announce` setting.
And a default value is set if the `netplay.nickname` is empty.

I apologize for bothering the Batocera team.
Feel free to let me know if there are any parts that need to be revised.


bulzipke:
Sorry about that. I mistakenly created a conflict with another branch, so I'll request a PR again.

PR #1869, opened 2025-02-18
Fontface Artifact Fix

bulzipke:
image
When the game is launched and then exited back to ES, the font of the popupkeyboard becomes distorted.
Vertical lines appear on the letters Q, A, and S.

This issue occurs when built with GLES20.

Original issue from: https://github.com/EmuELEC/EmuELEC/issues/878
Original PR: https://github.com/EmuELEC/emuelec-emulationstation/pull/100

### DIFF
--- a/es-core/src/renderers/Renderer_GLES20.cpp
+++ b/es-core/src/renderers/Renderer_GLES20.cpp
@@ -671,18 +671,25 @@ namespace Renderer
 		// Regular GL_ALPHA textures are black + alpha in shaders
 		// Create a GL_LUMINANCE_ALPHA texture instead so its white + alpha
 		
-		if (type == GL_LUMINANCE_ALPHA && _data != nullptr)
+		if (type == GL_LUMINANCE_ALPHA && _data == nullptr)
 		{
-			uint8_t* a_data  = (uint8_t*)_data;
+			uint8_t* la_data = new uint8_t[_width * _height * 2];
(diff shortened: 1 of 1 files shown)


fabricecaruso:
This PR is incredible !!!!
- I don't have artefacts with ES without this fix !
- It crashes, because the temp array is not correctly dimensionned !!!!
- The first "if" creates the artifacts meant to be fixed...

Incredible !

bulzipke:
> This PR is incredible !!!!
> 
> * I don't have artefacts with ES without this fix !
> * It crashes, because the temp array is not correctly dimensionned !!!!
> * The first "if" creates the artifacts meant to be fixed...
> 
> Incredible !

I really appreciate your insights, especially since you're the one who originally fixed this issue [link](https://github.com/EmuELEC/EmuELEC/issues/878#issuecomment-1086619995). It’s clear that you have a deep understanding of this part.

In my case, I'm running on a Linux h700, and I encountered an issue when returning to ES after exiting a game. This fix resolved that problem on my end.

That said, I'll definitely take a closer look at the crash you pointed out with the temp array. If you have any suggestions on improving this further, I'd be happy to refine the code together.

fabricecaruso:
Probably, but uint8_t[_width * _height * 2] for GL_RGBA is not enough : it requires 4 bytes per pixel, so it crashes with most video cards.
I'll provide a fix myself

fabricecaruso:
done

PR #1876, opened 2025-02-26
[NetPlay] Add Lan game search in lobby

bulzipke:
image

Support browsing RetroArch games in the lobby within the same network.
These games are added to the LAN GAMES group, while games retrieved from lobby.libretro.com are added to the ONLINE GAMES group.

### DIFF
--- a/es-app/src/guis/GuiNetPlay.cpp
+++ b/es-app/src/guis/GuiNetPlay.cpp
@@ -2,6 +2,9 @@
 #include "Window.h"
 #include <string>
 #include <future>
+#include <arpa/inet.h>
+#include <unistd.h>
+#include <fcntl.h>
 #include "Log.h"
 #include "Settings.h"
 #include "SystemConf.h"
@@ -153,7 +156,8 @@ GuiNetPlay::GuiNetPlay(Window* window)
 	mBusyAnim(window),
 	mBackground(window, ":/frame.png"),
 	mGrid(window, Vector2i(1, 3)),
-	mList(nullptr)
+	mList(nullptr),
+	mLanLobbySocket(-1)
 {	
 	addChild(&mBackground);
 	addChild(&mGrid);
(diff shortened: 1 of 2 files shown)


Darknior:
Really an excellent idea :)
@bulzipke do you think there is an available solution for network emulators that are not libretro ?
Like XBOX, or Nintendo DS or some other ... that have an network option but that we can't easily use in Batocera ...
I know it will be impossible to list games because there is no place for it, but maybe adding an option in ES to join a game with local IP of the server.

bulzipke:
> Really an excellent idea :) @bulzipke do you think there is an available solution for network emulators that are not libretro ? Like XBOX, or Nintendo DS or some other ... that have an network option but that we can't easily use in Batocera ... I know it will be impossible to list games because there is no place for it, but maybe adding an option in ES to join a game with local IP of the server.

That sounds like an exciting idea!
It would be amazing if standalone emulators could also be listed in Batocera's Netplay lobby. 

To achieve this, a central server would be needed to detect whether a standalone emulator is running in host mode—similar to how RetroArch’s lobby server works. 

Batocera could periodically detect if a specific port is listening in the background and report the running emulator, game, CRC32, and IP information to the central server. 

However, I’m not sure if all standalone emulators support CLI-based connections for joining games as a client. 

Since my main expertise is in server programming, setting up a relay server wouldn’t be difficult.
This is an interesting topic, and I'd love to discuss it further!

Feel free to share your thoughts—(perhaps I misunderstood your reply)

Darknior:
@bulzipke thanks for your answer :)
Yes coding a lobby will be a pain ... but adding only an option in ES to give the IP of the erveur is possible.
And if you know how to scan ports in local network in bash or python, to auto show servers in lobby, with a fake trick ... it will be fantastice, it is a great idea.
But for sure, first you must read emulator doc to see if it is possible for some or not ...
i think it is possible for xbox, and maybe Nintendo DS ... maybe also GC and Wii with Dolphin ?
Standalone Dolphin is for now the best version ...

fabricecaruso:
wonderfull. This does not compile with Windows and breaks Retrobat.....

bulzipke:
> wonderfull. This does not compile with Windows and breaks Retrobat.....

I see, thanks for the report. I'll check the issue and try to fix the Windows compilation problem. Regarding Retrobat, is there a separate repository for it, or is there any specific environment where I can test it? That information would help me debug the issue more effectively.

bulzipke:
> wonderfull. This does not compile with Windows and breaks Retrobat.....

I just realized that Retrobat is actually the Windows build repository of Batocera. When I worked on this, I was only considering Batocera.

Would you prefer if I add an #ifdef condition to make this change apply only to Batocera, or would it be better for me to adapt it so that it also works on the Windows version?

PR #1881, opened 2025-03-01
winsock not for linux

dmanlfc:


fabricecaruso:
Oh. Sorry. 
Thx for the fix

PR #1887, opened 2025-03-05
[NetPlay] Games always launch as Netplay Host, if public announce is set to ON ( fix bulzipke )

fabricecaruso:


bulzipke:
The idea behind this change was to ensure that games automatically function as a host when public announce is enabled, making it easier for players to host sessions without additional setup.

If there's a specific concern with this approach, I can adjust it rather than completely removing it. I'm open to suggestions on how to refine the implementation.

Would it help if I made this behavior another configurable so that users can enable or disable it as needed?

Also, if the issue is compatibility, I can test and ensure that it doesn’t break anything on other platforms.

fabricecaruso:
Probably, but the idea is not good.
Hosting a netplay game must be related to a voluntary user action.
How wants to host games everytimes he starts a game ? Probably nobody....

bulzipke:
When I was a kid, I used to go to the arcade, and one of the most exciting moments was when a stranger joined my game as Player 2—whether to challenge me or to help me clear a co-op game. (Back then, there was no way to set a password or block it with an option, haha.)

My intention with this feature was to recreate that kind of experience—where hosting a session is seamless and effortless. I felt that holding the A button to host might be an extra hurdle for more casual users.

Also, while RetroArch’s lobby system is great, there aren’t many active hosts. Unless you specifically plan a session with a friend, there are very few available rooms to join. That’s why I thought Batocera’s larger user base could help bring back that nostalgic experience.

Additionally, Batocera makes it easy to set up MITM relay functionality, allowing users to bypass complicated firewall or router configurations in advance. This ensures that players can connect smoothly without dealing with network setup issues.

Of course, this feature should always remain optional.

I really appreciate the time and effort you put into this project. If you have any suggestions, I’d be happy to collaborate and adjust the implementation accordingly.

fabricecaruso:
I have no opposition if you add an option in the netplay settings to do that, but this option should be disabled by default.

PR #1901, opened 2025-04-03
Update emulationstation2.po

Tartifless:
Correction french po


Tartifless:
@nadenislamarre : fix some french translations that meant nothing

PR #1923, opened 2025-05-02
WIP - Integrate zaparoo in the UI

asturur:
I m working at a possible integration between zaparoo and emulation station from the UI.
So far in zaparoo 2.3.1 ui support is obtained with the usage of a couple of scripts:

game select
https://github.com/ZaparooProject/zaparoo-core/blob/main/cmd/batocera/scripts/configs/emulationstation/scripts/game-selected/zaparoo_game_select.sh

and zaparoo writegame
https://github.com/ZaparooProject/zaparoo-core/blob/main/cmd/batocera/scripts/zaparoo_write_game.sh

and a multimedia_key.conf to activate writing with a button press
https://github.com/ZaparooProject/zaparoo-core/blob/main/cmd/batocera/scripts/configs/multimedia_keys.conf

While this method works it has a series of drawbacks like installation complexity and poor feedback.

This PR aims at staring a discussion for handling this directly in the UI.

I m still figuring out how to compile my code, so this may not work at all, i need to setup an sd card with a distro that allows me to compile on the rasp 4, unless someone has a guide to compile on apple silicon for the rasp architecture.


fabricecaruso:
Is this still relevant?

es-app/src/ApiSystem.cpp
```diff
@@ -2307,3 +2307,19 @@ bool ApiSystem::enableService(std::string name, bool enable)
...
+
+	auto slines = executeEnumerationScript("batocera-services status " + serviceName);
+	auto sline = slines[0];
+	return sline.compare("started") == 0;
+}
+
+bool ApiSystem::writeZaparooCard(std::string name)
+{
+	return executeScript("/userdata/system/zaparoo -write \"" + name + "\"");
+}
```
asturur: @fabricecaruso since you seem knowleageable in the matter, and since this emulation-station from what i understand from the makefile also powers retrobat, would you suggest that this code should be written platform independent? And if yes, could you help me doing so?

fabricecaruso: >  If the service is not up it won't connect, but is going to be fast enough to do every time we open the menu?

You're right, if the service is not running, ES would hang to the default timeouts which is 2 seconds by default. We'd have to reduce timeout values ( to 100ms ? )

> Also in general http request is something i see as async

Can you explain, why would it be async. The http response ( 200 ? ) would not indicate that it's finished and OK ?

es-app/src/ZaparooSupport.cpp
```diff
@@ -0,0 +1,102 @@
...
+    options.customHeaders = {"Content-Type: application/json"};
+    options.dataToPost = "{\"jsonrpc\": \"2.0\",\"id\": \"" + uuid + "\",\"method\": \"readers.write\",\"params\": {\"text\": \"" + name + "\"}}";
+    HttpReq req("http://127.0.0.1:7497/api/v0.1", &options);
+    if (req.wait()) {
+        // may need to inspect the response here
+        return true;
+    } else {
+        return false;
+    }
+}
```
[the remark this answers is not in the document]
asturur: I m doing more tests tomorrow, for now everything works.
I m not going to change more code outside maybe the return of this method to detect errors in writing.
So please if you can have a re review and guide me for the translation strings of which i ignore the contribution guidelines

es-app/src/ZaparooSupport.cpp
```diff
@@ -0,0 +1,102 @@
...
+    if (req.wait()) {
+        std::string data = req.getContent();
+        return data.compare("OK") == 0;
+    }
+    return false;
+}
+
+bool Zaparoo::writeZaparooCard(std::string name)
+{
+    HttpReqOptions options = getHttpOptionsZaparoo();
```
fabricecaruso: you should not set a low timeout for the write method. only the "health" method requires a low timeout.

es-app/src/ZaparooSupport.cpp
```diff
@@ -0,0 +1,102 @@
...
+{
+	if (Zaparoo::sInstance == nullptr)
+		Zaparoo::sInstance = new Zaparoo();
+
+	return Zaparoo::sInstance;
+}
+
+bool Zaparoo::executeScript(const std::string command)
+{	
+
```
fabricecaruso: Don't write a call to system().
On Windows, that pops a cmd window. 
Instead, call ApiSystem::executeScript() ( which is overriden for windows )

es-app/src/ZaparooSupport.cpp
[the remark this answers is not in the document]
fabricecaruso: > moved those 2 zaparoo functions in their own module.

That's great

es-app/src/ZaparooSupport.cpp
asturur: the options are new for every write, so for this request i don't intend to change the timeout since we already stated with the above check that zaparoo is up.

es-app/src/ZaparooSupport.cpp
```diff
@@ -0,0 +1,102 @@
...
+    HttpReqOptions options = getHttpOptionsZaparoo();
+    std::string uuid = generateUUID();
+    LOG(LogInfo) << "Zaparoo::writeZaparooCard -> " << uuid;
+    options.customHeaders = {"Content-Type: application/json"};
+    options.dataToPost = "{\"jsonrpc\": \"2.0\",\"id\": \"" + uuid + "\",\"method\": \"readers.write\",\"params\": {\"text\": \"" + name + "\"}}";
+    HttpReq req("http://127.0.0.1:7497/api/v0.1", &options);
+    if (req.wait()) {
+        // may need to inspect the response here
+        return true;
+    } else {
```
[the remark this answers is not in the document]
asturur: Yes this was before `return req.wait();` but i was always getting a false because a zaparoo bug, while investigating if i had to inspect the actual json i discovered that zaparoo was returning an OK 500 because a bag.

es-core/src/HttpReq.cpp
```diff
@@ -191,8 +191,8 @@ void HttpReq::performRequest(const std::string& url, HttpReqOptions* options)
 	// Ignore expired SSL certificates
 	curl_easy_setopt(mHandle, CURLOPT_SSL_VERIFYPEER, 0L);
 
-	//set curl to handle redirects
-	err = curl_easy_setopt(mHandle, CURLOPT_CONNECTTIMEOUT, 10L);
+	//set curl to connection timeout
+	err = curl_easy_setopt(mHandle, CURLOPT_CONNECTTIMEOUT, options != nullptr ? options->connectTimeout : 10L);
```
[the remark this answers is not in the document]
asturur: yes ms seems better. 1 second seemed too much to me.

asturur:
Ok everything works as expected.
@fabricecaruso i would need help to add the translation strings in.
Every time i compile all the language file gets reorganized and create enormous diffs.
How do i proceed to add my lanugage strings in in english.
Do i have to provide the rest or you have a community or a service to get the translation strings in the other languages?

PR #1929, opened 2025-05-09
Add user community relay server feature with SEOUL as the first preset

bulzipke:
Batocera currently provides RetroArch’s official relay servers as options for netplay.
This update adds support for user community relay servers, allowing additional community-hosted relays to be included.
I have added a relay server for Korean players (SEOUL) as the first community option, and I hope this opens the door for more countries to add their own in the future.

from https://github.com/knulli-cfw/batocera-emulationstation/pull/51

### DIFF
--- a/es-app/src/guis/GuiNetPlaySettings.cpp
+++ b/es-app/src/guis/GuiNetPlaySettings.cpp
@@ -6,10 +6,9 @@
 #include "GuiHashStart.h"
 #include <unordered_map>
 
-static std::unordered_map<std::string, std::string> communityRelayServers =
-{
-	{"seoul",		"138.2.119.137"},
-};
+#define fake_gettext_seoul _("SEOUL")
+
+static std::vector<std::pair<std::string, std::string>> communityRelayServers;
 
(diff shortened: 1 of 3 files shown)


fabricecaruso:
Ain't convinced it's a good way to hardcode non official servers in the code.
Maybe store this in a file instead so it's easier to update for users that can't read code ?

bulzipke:
Thank you for the suggestion!

I’m wondering if Batocera already has a similar configuration file. I only know about `batocera.conf` at the moment, but what do you think about managing it in a JSON format like the example below?

```json
[
  {
    "name": "SEOUL",
    "url": "relay.seoul.community:55355"
  },
  {
    "name": "PARIS",
    "url": "relay.paris.community:55355"
  }
]
```


Alternatively, I’m also considering fetching this JSON from a GitHub repo or another remote URL.
If Batocera already supports loading external configuration in a similar way, I’d appreciate it if you could let me know!

fabricecaruso:
as most files used by EmulationStation are xml files ( arcaderoms.xml, gamesdb.xml, es_settings, e_systems... ) I think it would be preferable to stay with xml.

bulzipke:
I’ve implemented dynamic loading of community relay servers from an external XML file.  
Please let me know if you have any further suggestions!

PR #1930, opened 2025-05-09
[NetPlay] Add automatic hotspot creation feature

bulzipke:
image

**Short Summary**  
Added Adhoc (Hotspot) feature for local Netplay with nearby players.  
Users can now play Netplay together without going through a router.
When the `AUTOMATICALLY USE HOTSPOT FOR LOCAL NETPLAY` option is enabled, a hotspot will be created when starting a game if Wi-Fi is not available, and the Netplay menu will search for available hotspots.

**Detailed Description**
1. **es-app/src/ApiSystem.cpp**
   - Added logic to retrieve the IP address of the current Wi-Fi route.
   - The existing queryIPAddress function does not ensure that the network is fully ready.
   - This update ensures that LAN games can be reliably discovered via broadcast immediately after connecting to a hotspot.

2. **es-app/src/guis/GuiGameOptions.cpp**
3. **es-app/src/views/ViewController.cpp**
   - If the hotspot option is enabled, the system will attempt to create a hotspot and act as a host even without an internet connection.

4. **es-app/src/guis/GuiNetPlay.cpp**
   - In hotspot mode, `mLobbyRequest` consistently fails and causes unnecessary error dialogs.
   - To address this, if LAN games exist (`mList` is not empty), the error is ignored.
   - Additionally, a short delay (100ms) was added before error checking to give LAN broadcasts time to complete.
   - The `populateFromLan` function was also updated to handle list changes (`changed`) internally.

5. **es-app/src/guis/GuiNetPlaySettings.cpp**
   - Added the hotspot option toggle.
   - It is hidden on Windows, where hotspot-related functionality is not implemented.

6. **es-app/src/views/SystemView.cpp**
   - When there is no active Wi-Fi connection and the hotspot option is enabled, pressing the Netplay button will initiate a scan for nearby hotspot APs (up to 20 seconds).
   - If Wi-Fi is already connected or the option is disabled, it behaves as before.

**Related Issue**  
https://github.com/batocera-linux/batocera.linux/pull/14075

from https://github.com/knulli-cfw/batocera-emulationstation/pull/39

### DIFF
--- a/es-app/src/ApiSystem.cpp
+++ b/es-app/src/ApiSystem.cpp
@@ -490,6 +490,13 @@ std::string ApiSystem::getIpAddress()
 	return result;
 }
 
+bool ApiSystem::isWifiAPModeSupported()
+{
+	LOG(LogDebug) << "ApiSystem::isWifiAPModeSupported";
+
+	return executeScript("batocera-wifi has_ap_mode");
+}
+
 bool ApiSystem::enableBluetooth()
 {
 	return executeScript("batocera-bluetooth enable 2>&1 >/dev/null");
@@ -1563,7 +1570,7 @@ void ApiSystem::scanWifiNetworks()
 
 std::string ApiSystem::getWifiRoute()
 {
-	std::vector<std::string> result = executeEnumerationScript("batocera-wifi getroute");
+	std::vector<std::string> result = executeEnumerationScript("batocera-wifi get_route");
 
 	if (result.empty() || result[0].empty())
 		return "NOT CONNECTED";
--- a/es-app/src/ApiSystem.cpp
+++ b/es-app/src/ApiSystem.cpp
@@ -1556,6 +1556,21 @@ std::vector<std::string> ApiSystem::getWifiNetworks(bool scan)
 	return executeEnumerationScript(scan ? "batocera-wifi scanlist" : "batocera-wifi list");
 }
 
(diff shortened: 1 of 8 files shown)


spycat88:
Just a suggestion but I would at implementing an API that detects whether a device can actually act as hotspot host as not all Linux drivers support host mode, if not then the option to host should be hidden. (done this in ROCKNIX recently)

bulzipke:
> Just a suggestion but I would at implementing an API that detects whether a device can actually act as hotspot host as not all Linux drivers support host mode, if not then the option to host should be hidden. (done this in ROCKNIX recently)

Thanks for the suggestion! That makes a lot of sense — hiding the option when host mode isn’t supported would improve usability and avoid confusion. I’ll look into how ROCKNIX handles this and see if something similar can be implemented here as well.

spycat88:
https://github.com/ROCKNIX/emulationstation-next/commit/4608a013eaf266190c3f37f9da70f1a19e7a9ece

thats the change we used, as well as a helper that just consists of the following

```
has_ap_mode() {
  PHY=$(basename "$(readlink /sys/class/net/${WIFI_DEV}/phy80211)")
  if iw phy "${PHY}" info | sed -n '/Supported interface modes/,/^[^*]*$/s|^.*[*] ||p' | grep -q 'AP'; then
    exit 0
  else
    exit 1
  fi
}
```

PR #1945, opened 2025-07-07
[CHEEVOS MENU] Add unofficial achievements switch

Tartifless:
This is used by dolphin and pcsx2 for example, it can be used to unlock unofficial achievements


Tartifless:
@nadenislamarre can you check this one ?

Tartifless:
Retrobat code (EL) is ready, if not ok for batocera i can make it windows only.
But batocera also has dolphin and pcsx2 cheevos, so this option can be useful also.

Tartifless:
@lbrpdx : all information requested sent on github.

It can be enabled in configgenfor duckstation, pcsx2, dolphin and ppsspp that have the option in the achievements section.

PR #1946, opened 2025-07-11
Update Swedish emulationstation2.po

yeager:


yeager:
Sorry, Will use https://translations.batocera.org/ instead

PR #1950, opened 2025-07-25
Fix script update-gamelists

Tartifless:
@fabricecaruso Can you check this one ?


Tartifless:
@dmanlfc : can you have a look ? The event trigger for update gamelist was wrongly placed, i placed it after the YES push and it's now working correctly

PR #1955, opened 2025-08-04
Switch to "fast" default transition

lbrpdx:
Would you be OK with defaulting to "fast" slide transition for launching games? It makes the whole experience feel snappier to me.
Please chime in if you disagree @fabricecaruso @dmanlfc @nadenislamarre @Tartifless


dmanlfc:
Sounds good to me

RetroBoyCN:
For devices with weaker performance such as the S905 chip, it is not friendly, with obvious lag and poor smoothness.

PR #1967, opened 2025-09-04
Added disc free space for auto mounted disks /media/diskname

SirRickster:
I have an additional SSD automatically mounted on /media where I symlink some of the roms folders on it. I find it useful to know how much free space there is left.
It can be useful for someone else.


fabricecaruso:
As it's a linux only PR, I let @nadenislamarre and @lbrpdx decide

PR #1968, opened 2025-09-04
Added: Launch filemanager from GUI

SirRickster:
New entry to launch Filemanager from GUI on developer tools menu


crcerror:
@lbrpdx meanwhile you can indeed use a mouse only system to use the filemanager. For wine games I use the FM to extract the game to create the autorun and to squash it and to symlink the default prefix.

But I think it use more usefull to add a simple shell script in ports instead. That is enough for the purpose I guess. Just call /usr/bin/filemanager

PR #1994, opened 2025-11-03
reduce the amount of display rendering to save battery

dmanlfc:
- idle FPS drops from 60fps to <10fps

### DIFF
--- a/es-app/src/main.cpp
+++ b/es-app/src/main.cpp
@@ -637,7 +637,7 @@ int main(int argc, char* argv[])
 	else
 		AudioManager::getInstance()->playRandomMusic();
 
-
+// This entire block may now be redundant and can be safely removed.
 #ifdef WIN32	
 	DWORD displayFrequency = 60;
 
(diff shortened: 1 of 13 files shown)


fabricecaruso:
Branch does not compile. 
Utils::Platform::ExitRequest::user does not exist

fabricecaruso:
Just tested it by fixing myself the code locally so it can compile.

It works very badly and makes the animated themes unusable. Try with Cinematics DX theme !!

SystemView extras are not resolved by isAnimating() method as SystemView has no children ( You'd have to override isAnimating() on SystemView and parse current system extras ), so getIdealTimeout() falls back to the ms_until_next_minute value -> The screen refreshes only every minute.

Also : 

- If an animation is active, forcing it to 30FPS gives the impression we are running on a slow machine.
  -> There should be no timeout at all when an animation is active.
- I don't understand the interest of ms_until_next_minute or the 2 seconds wait fallback. Setting a maximum value to 1 second will reduce the power enough.
- We can't disable VSYNC anymore to unleak the FPS, and test ES without any limit. ( it's usefull for optimization )

PR #2010, opened 2025-11-14
feat: keyboardtopads gui

nadenislamarre:


nadenislamarre:
to be merged once the python part will be finished.

PR #2013, opened 2025-11-18
debug-boot-mode switch [WIP]

Mikhailzrick:
Adds toggle switch for debug boot mode in batocera. In the front end dev menu, near the logging which makes sense to me.

Since initramfs is very early in boot process and we want to be able to skip using a saved overlay. Uses toggle to call S01bootdebug enable | disable which creates the flag file in /boot/boot/bootmode.debug

That way we can easily test -f the file in initramfs

https://github.com/batocera-linux/batocera.linux/pull/14816

### DIFF
--- a/es-app/src/guis/GuiMenu.cpp
+++ b/es-app/src/guis/GuiMenu.cpp
@@ -61,7 +61,7 @@
 #include "Gamelist.h"
 #include "TextToSpeech.h"
 #include "Paths.h"
-#include <set> 
+#include <set>
 
 #if !WIN32
 #include <vector>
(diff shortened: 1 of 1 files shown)


fabricecaruso:
Please : 
- Set your PR as Draft as long as it's not definitive.
- See the files changed in the PR : There are so many formatting differences in GuiMenu that we can't even see the new code. Please don't touch formatting. The files changed Tab should only show what you really changed.

Mikhailzrick:
> Please :
> 
> * Set your PR as Draft as long as it's not definitive.
> * See the files changed in the PR : There are so many formatting differences in GuiMenu that we can't even see the new code. Please don't touch formatting. The files changed Tab should only show what you really changed.

I have Kate auto remove empty lines and trialing spaces and such. So that’s why that’s happening. Batocera-emulationstation is really the only thing that causes issues with because there’s so many, but I can turn it off for ES PR’s. 

I am currently waiting on Susan because there’s a few ways we can go about this between ES <-> batocera.

fabricecaruso:
It's almost impossible in the current state to read what you changed in your PR because of the formatting changes.

PR #2022, opened 2025-11-29
handle in the gui messages when usb / sdcards are inserted etc

dmanlfc:
this change provide gui updates for when storage devices are added
this works hand-in-hand with the new batocera-storage-manager which handles device insertion, ejection events for usb & sdcards as well as storage tasks.

### DIFF
--- a/es-app/src/ApiSystem.cpp
+++ b/es-app/src/ApiSystem.cpp
@@ -2823,3 +2823,26 @@ bool ApiSystem::enableService(std::string name, bool enable)
 	
 	return res;
 }
+
+std::vector<std::string> ApiSystem::getEjectableDrives()
+{
+    return executeEnumerationScript("batocera-storage-manager list_ejectable");
+}
+
+bool ApiSystem::ejectDrive(const std::string& mountPath)
+{
+    std::string cmd = "batocera-storage-manager eject \"" + mountPath + "\"";
+    return executeScript(cmd);
+}
+
+bool ApiSystem::mergeDrive(const std::string& mountPath)
+{
+    std::string cmd = "batocera-storage-manager merge \"" + mountPath + "\"";
+    return executeScript(cmd);
+}
+
+bool ApiSystem::prepareDrive(const std::string& device, const std::string& fsType)
+{
+    std::string cmd = "batocera-storage-manager format \"" + device + "\" \"" + fsType + "\"";
+    return executeScript(cmd);
+}
--- a/es-core/src/Window.cpp
+++ b/es-core/src/Window.cpp
@@ -25,6 +25,8 @@
 #include "Splash.h"
 #include "PowerSaver.h"
 #include "renderers/Renderer.h"
+#include <thread>
+#include "../es-app/src/ApiSystem.h"
 
 #if WIN32
 #include <SDL_syswm.h>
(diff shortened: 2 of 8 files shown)


fabricecaruso:
I really don't like the idea of checking some strange files appear by some magical in each Window::update call and check some unknown external code wrote a /tmp/es-hotplug-event or /tmp/es-pending-storage file !?!

I hate this way to do things ! 

The good way would be to work with events ( or something that looks like events when the api has no real events ) !

Instead, either :
- Stop writing external scripts for this kind of things and write true C++ code, then handle udev_monitor_receive_device in a thead to throw an event.... The getEjectableDrives and ejectDrive methods can be written easily in true C++ code. The monitoring requires a thread on linux but is easy to write too.  We now have IA to help, I'm sure ChatGPT or Claude would help write beautiful C++ code that works !
I'll never understand why, you guys, always prefer write scripts that can be shared with no-one, instead of writing C++ code that can be reused by other distrubutions, especially when the script doesn't do anything very complicated and is very easy to write in C++. There are so many things with batocera that require scripts where is so easy to write things in C++ ( bluetooth, wifi, format, overclock and many others !!! )

- If you really wanna keep external scripts, then make a REST api in HttpServerThread. The you can call a REST api and you can manage it as if it was an event. Also, don't hardcode paths when you call a script ( /usr/bin/batocera-storage-manager : NO -> why /usr/bin ???? ). 

Honestly, I personnally hate this external scripts and checking the existance of files everywhere. You, do this everytime and it has become uncomprehensible to understand how things work because there are so many of them going in all directions.

Also : why so scripts are always named "batocera-something" ??? When someone want to port this ( emuelec, retrobat... ), we need : 
- To port scripts. If it was true c++ code, any other linux distribution or standalone ES will work immediately ( except windows but it's not a problem, I'll manage ).
- Retrobat or EmuElec or other distributions have to name their scripts "batocera-something".

fabricecaruso:
> Wow! Someone having a bad day? 
Mmm, rather a pretty bad year. and a very bad week...

>Is this not `batocera-emulationstation` ? 
Yes, but it's now used by many other distributions. Mainly Retrobat ! But also by many linux distros.
I try to care about the fact that the code, as it's open souce, can be used by everybody.

>The code is also wrapped in #ifdef BATOCERA
I know, and for specific things... Why not... But when it's about system things like this, it should not be limited to batocera, but to the OS ( #if defined(__linux__) ). 

I'm not against having external scripts, but I really think we should avoid them when what the script do is "operating system" related ( I mean it's a linux/windows/macos code that is not really related to batocera, but to the operating system under batocera ).

In fact, what really bothered me about your PR is not the script thing ( it's a whole ) but the fact that you check the existance of 2 files, when it's impossible to understand where they come from by reading the c++ code. Regarding this kind of thing, I am really very opposed to it.

fabricecaruso:
> your concern is valid and I will pivot to instead of ES pulling the details from a /tmp file I will move to using the HTTP api and push notifications and the refresh to the gui.

Yes please, that would be a far better way to do this

dmanlfc:
> > your concern is valid and I will pivot to instead of ES pulling the details from a /tmp file I will move to using the HTTP api and push notifications and the refresh to the gui.
> 
> Yes please, that would be a far better way to do this

OK done, it's working well.
Let me know what you think...

es-app/src/ApiSystem.cpp
```diff
@@ -2823,3 +2823,14 @@ bool ApiSystem::enableService(std::string name, bool enable)
 	
 	return res;
 }
+
+std::vector<std::string> ApiSystem::getEjectableDrives()
+{
+    return executeEnumerationScript("/usr/bin/batocera-storage-manager list_ejectable");
```
fabricecaruso: please don't hardcode paths. It should work without "/usr/bin/"

fabricecaruso:
I commented where you have fixes to do

es-core/src/Window.cpp
```diff
@@ -1427,3 +1428,55 @@ bool Window::processMouseButton(int button, bool down, int x, int y)
...
+			_("YES, FORMAT & MERGE"), [this, argument] {
+				
+                this->displayNotificationMessage(_("Starting format process..."));
+
+				std::thread([this, argument]() {
+                    // Get the user's preferred filesystem (Defaults to exfat)
+                    std::string fsType = SystemConf::getInstance()->get("system.external_disk_format");
+                    if (fsType.empty()) fsType = "btrfs";
+
+					int ret = std::system(("/usr/bin/batocera-storage-manager format \"" + argument + "\" \"" + fsType + "\"").c_str());
```
fabricecaruso: See my previous comment : no hardcoded paths + move this to ApiSystem

PR #2028, opened 2025-12-06
Integrate Zaparoo in the emulation station UI

asturur:
THIS IS THE SAME PR BUT REBASED ON MASTER

I m working at a possible integration between zaparoo and emulation station from the UI.
So far in zaparoo 2.3.1 ui support is obtained with the usage of a couple of scripts:

game select
https://github.com/ZaparooProject/zaparoo-core/blob/main/cmd/batocera/scripts/configs/emulationstation/scripts/game-selected/zaparoo_game_select.sh

and zaparoo writegame
https://github.com/ZaparooProject/zaparoo-core/blob/main/cmd/batocera/scripts/zaparoo_write_game.sh

and a multimedia_key.conf to activate writing with a button press
https://github.com/ZaparooProject/zaparoo-core/blob/main/cmd/batocera/scripts/configs/multimedia_keys.conf

While this method works it has a series of drawbacks like installation complexity and poor feedback.

This PR aims at staring a discussion for handling this directly in the UI.

This PR is based on the commit of emulation-station that was used for batocera v42 becuase if i use latest master i cannot really compile easily because of some latest additions that require the flag BATOCERA to be set.
I took this commit as a reference one for testing on batocera 42, but the diffs are so small that it shouldn't be an issue if it pass QA testing to build it on top of latest master

## How is tested.

I compile it on macos apple silicon with the help of a docker image, i get a generic linux arm 64 build, then i copy it inside my rpi4 live installation, i mount boot as rw, i put the compiled binary inside /usr/bin, i do a save-overlay and i reboot and i test manually that it works

## PR in action

The code detects if the zaparoo service returns 'started' when checked for status, and if so adds an entry in the menu.

image1

Once selected you get a dialog

image2

And then a small waiting state for you to actually put the card on/in the reader

image3

And if all goes good a notice that all went good

image4

Soemetimes instead you get an error

image5

## Translation strings for reference

```
msgid "WRITE GAME ON NFC TAG"
msgstr "WRITE GAME ON NFC TAG"

#, c-format
msgid ""
"IN ORDER TO WRITE THE LAUNCH COMMAND FOR\n"
"'%s'\n"
"PRESS THE WRITE BUTTON AND THEN PLACE AN NFC TAG ON THE WRITER"
msgstr ""

msgid "WRITE"
msgstr ""

msgid "PLACE A TAG ON THE WRITER ..."
msgstr ""

msgid "THE TAG WAS WRITTEN SUCCESSFULLY"
msgstr ""
```


asturur:
I will provide more translations asking friends, but for now this PR is done for me ( outside  other requests for changes )

PR #2035, opened 2025-12-10
Add an optional "sort name" that gets precedence over (scraped) game …

lbrpdx:
…name


es-app/src/FileSorts.cpp
```diff
@@ -71,12 +71,25 @@ namespace FileSorts
 		mSortTypes.push_back(SortType(RELEASEDATE_SYSTEM_DESCENDING, &compareReleaseYearSystem, false, _("RELEASE YEAR, SYSTEM, DESCENDING"), _U("\uF161 ")));
 	}
 
+	static std::string getSortName(const FileData* file)
```
fabricecaruso: You could refactor this into a FileData::getSortName() method in order this can be reused.

lbrpdx:
Refactoring done.

PR #2051, opened 2026-01-02
system view : add option to filter by hardware category

aderumier:
When enabled,left|right or up|down is switch to next hardware category and only diplay systems from this category.

Ordering by date, manufacter,name,.. is still working inside a filtered hardware categories

Some themes like
https://github.com/BobMoraneRetro/one4all4one-coinopsfriends-bobmorane

already provide nice icons to diplay the current selected categories.  (without theme support, are still a notification indication when the categorie is changed)


aderumier:
ping @fabricecaruso   what do you think about this patch ?

Darknior:
Excellent idea :D

PR #2083, opened 2026-02-27
add option to disable game scan at startup

DevNamedZed:
- Adds SQLite game cache (gamedatabase.db) that persists discovered games across restarts                                                        
- New "Skip game scan at startup" option in Developer Settings — loads from cache instead of filesystem walk                                   
- "Scan for new games" button in Game Settings for manual rescan when skip is enabled                                                            
- Defaults to off — zero behavior change unless opted in, graceful fallback if DB is missing     

Tested on mint machine with `-DBATOCERA=OFF` and a few homebrew roms. Confirmed DB is created and log messages stating DB is being used vs scanning. Tested add/removing roms, confirmed scan does not happen when setting is turned on and manual scanning still work, updates the db


DevNamedZed:
> I just tested this on Windows.
> 
> 1/ I don't really see any performance improvement compared to the PARSE GAMELISTS ONLY option. Here are my bencharks for 9251 games: PARSE GAMELISTS ONLY 2026-03-16 22:13:56 TIME 1,115 2026-03-16 22:14:20 TIME 1,249 2026-03-16 22:14:39 TIME 1,713
> 
> SKIP GAME SCAN AT STARTUP 2026-03-16 22:13:19 TIME 1,609 2026-03-16 22:15:16 TIME 1,173 2026-03-16 22:15:34 TIME 1,404 So it's almost the same performance.
> 
> Did you benchmark it ?
> 
> 2/ If the PARSE GAMELISTS ONLY is unchecked, then there are no images at all for games. In the Metadata class, mRelativeTo is null so it the full path is never resolved. It's not mergeable in the current state because of this problem with the games medias.
> 
> Also, the tables structures are not dynamic (every field is declared), so if we need to make any evolution with metadatas, we need to maintain the database, too, and make code to upgrade the database. That complexifies maintenance.
> 
> I really wonder what's the interest of introducig such a thing.

For me, I have my games on a ZFS server. Before I did lots of optimization (switch from smb to nfs, add lots of ram for L1 arc, optane l2 arc driver, tweak the ZFS metadata cache size, 10gbe wired, etc) cold boot time used to take up to 30 minutes.

Testing my setup, with 125K games, I see a 10-30x performance improvement with a warm ZFS arc cache (1-2 seconds, compared with 10-30 seconds).  

ZFS had 36gb of metadata cached in memory, I rebooted my storage array to simulate a cold boot and ran just now. With the sqlite db it starts up instantly (1.5 seconds), without I get 94 seconds to scan. If my setup wasn't so nice it would be much much slower.

```
Cold NFS scan: 93,851ms (94s)
DB cache load: 1,451ms (1.5s)
```

I have an ally-x also that I run ES-DE. Not the same as this, but can also see a huge benefit from disable scan at startup. On that device I have my roms on a 2tb SD card, SD card is very very slow. It takes a good 15-30minutes everytime I start. This will be helpful for people with slow storage. 

Let me know if it's okay to move forward with this. I can address the media path issue and can change to store stuff as json blobs.

DevNamedZed:
> I did not test with games & gamelists stored on a network share. It's possible it makes a true difference.
> 
> I have no objections on your PR, but I need arguments and it has to work perfectly, and be easy to maintain ;-)
> 
> * First : You most solve the image problem - it's very strange you did not notice it ! It's because as you don't call loadFromXML, mRelativeTo is null and then relative paths can't be resolved.
> * I don't like the condition "skipScan = Settings::SkipGameScanAtStartup() && !Settings::ParseGamelistOnly();" -> If you check both options, SkipGameScanAtStartup does not apply, only ParseGamelistOnly does. It's not user friendly at all. How will understand ? Maybe this should be regrouped in a single combobox, something like "LOADING MODE : NORMAL / PARSE GAMELISTS ONLY / USE GAME DATABASE"
> 
> Also one of my main concern is that **there is no database upgrade code.** I mean : If we need to change the metadata structure - imagine we want to add one new metadata ( I added the "tags" metadata recently ). How do we do this ? There's no "automatic" way.
> 
> Faced with this question, I'm wondering about the relevance of the database structure you implemented. As it stands, each metadata field is a named column, which forces us to create new columns, and write code to update the structures each time we want to add something in the metadatas.
> 
> Maybe it should be managed - not with columns, but - with another "metadata" table. The `games` table would only have the bare minimum (id, system, path, type, and possibly name -> everything required that is not declared in the `MetaDataId` enum), and another table, `game_metadata` (game_id, key, value).
> 
> It's a bit more complicated to write load & save, but there would be no questions about future evolutions.
> 
> If you have other suggestions. But this question has to be answered !



> I did not test with games & gamelists stored on a network share. It's possible it makes a true difference.
> 
> I have no objections on your PR, but I need arguments and it has to work perfectly, and be easy to maintain ;-)
> 
> * First : You most solve the image problem - it's very strange you did not notice it ! It's because as you don't call loadFromXML, mRelativeTo is null and then relative paths can't be resolved.
> * I don't like the condition "skipScan = Settings::SkipGameScanAtStartup() && !Settings::ParseGamelistOnly();" -> If you check both options, SkipGameScanAtStartup does not apply, only ParseGamelistOnly does. It's not user friendly at all. How will understand ? Maybe this should be regrouped in a single combobox, something like "LOADING MODE : NORMAL / PARSE GAMELISTS ONLY / USE GAME DATABASE"
> 
> Also one of my main concern is that **there is no database upgrade code.** I mean : If we need to change the metadata structure - imagine we want to add one new metadata ( I added the "tags" metadata recently ). How do we do this ? There's no "automatic" way.
> 
> Faced with this question, I'm wondering about the relevance of the database structure you implemented. As it stands, each metadata field is a named column, which forces us to create new columns, and write code to update the structures each time we want to add something in the metadatas.
> 
> Maybe it should be managed - not with columns, but - with another "metadata" table. The `games` table would only have the bare minimum (id, system, path, type, and possibly name -> everything required that is not declared in the `MetaDataId` enum), and another table, `game_metadata` (game_id, key, value).
> 
> It's a bit more complicated to write load & save, but there would be no questions about future evolutions.
> 
> If you have other suggestions. But this question has to be answered !

Thank you so much for reviewing. I have a batocera machine connected to every TV in my house and really love it.

I updated so the metadata goes in a json blob. Reduced the number of columns to 6, use MetaDataDecl for fields to store in the blob. Probably no schema changes ever, new metadata fields added to gameDecls are automatically picked up.

  id INTEGER PRIMARY KEY AUTOINCREMENT,
  system TEXT NOT NULL,
  path TEXT NOT NULL,
  type INTEGER NOT NULL DEFAULT 0,
  metadata TEXT,
  last_synced TEXT DEFAULT CURRENT_TIMESTAMP,

If there is though, there is a table db_meta that tracks the schema version. If the version changes, we just drop the table and recreate. It will force a rescan. This is a transient cache for performance, its not the source of truth, so this should be okay I think

Fixed the image/media path issue — added setRelativeTo(system) so relative paths resolve correctly when loading from cache.

I will work on the dropdown, using a separate table for metadata, and other feedback later tonight. Please let me know if any other issues

DevNamedZed:
This will probably work great. From my testing, time was dominated with the multi disk content. I will test later tonight.

Is there a way to get it to write out the gamelist files when just a file name? 

https://github.com/batocera-linux/batocera-emulationstation/blob/master/es-app/src/Gamelist.cpp#L241

PR #2114, opened 2026-04-14
[WIN32] Fix dualsense detection once and for all

Tartifless:
Tested with RetroBat, fixes Dualsense detection when controller is already plugged in at RetroBat start


Tartifless:
@lbrpdx @dmanlfc @fabricecaruso : can one of you merge this ?

PR #2115, opened 2026-04-16
Startup speed

XenuIsWatching:
Two improvements to FileCache to reduce startup time when many systems
load in parallel via ThreadPool:

1. Upgrade std::mutex to std::shared_mutex. Read-only cache hits
   (exists, isDirectory, etc.) now use a shared_lock, allowing all
   threads to read concurrently with no blocking between them.

2. Move stat64/filesystem I/O out of the exclusive lock. Previously
   getCacheEntry() performed stat() while holding the write lock,
   serializing all threads through slow network I/O. Now statKey()
   does the I/O with no lock held; the exclusive lock is only acquired
   for the brief map insertion, with a double-check for races.

For absolute paths (the common case for ROM paths), getCanonicalPath was
calling exists() which performs a stat64 just to decide whether to call
getAbsolutePath() — but getAbsolutePath() on an already-absolute path is
a no-op anyway. Skip the stat entirely by checking isAbsolute() first.

Also add an early return when the path contains no '.' components, since
there is nothing to normalize. This mirrors the same optimization already
present in the Windows code path above.

Together these eliminate the per-ROM stat calls in removeMultiDiskContent
→ getCanonicalPath that were serializing startup on network-backed ROM
libraries.

### DIFF
--- a/es-core/src/utils/FileSystemUtil.cpp
+++ b/es-core/src/utils/FileSystemUtil.cpp
@@ -10,6 +10,7 @@
 #include <string.h>
 #include <algorithm>
 #include <set>
+#include <shared_mutex>
 
 #if defined(_WIN32)
 // because windows...
@@ -91,7 +92,7 @@ namespace Utils
 				if (!Settings::UseFileCache())
 					return;
 
-				std::unique_lock<std::mutex> guard(mFileCacheMutex);
+				std::unique_lock<std::shared_mutex> guard(mFileCacheMutex);
 
 				auto [it, inserted] = mFileCache.try_emplace(hashPath(key), data);
 				if (!inserted)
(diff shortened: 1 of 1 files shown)


fabricecaruso:
I agree with the use of the std::shared_mutex, which will be better for reading cache.
I'm not convinced by the code repetition you did deleting the getCacheEntry method -> The code is now is duplicated 5 times (only return conditions differs) !!!

XenuIsWatching:
> I agree with the use of the std::shared_mutex, which will be better for reading cache. I'm not convinced by the code repetition you did deleting the getCacheEntry method -> The code is now is duplicated 5 times (only return conditions differs) !!!

alright, i've fixed that, i also split that commit in to 2 as well here

fabricecaruso:
There are still many repetitions.

Wouldn't it be better if we refactor so we use lambdas.  Something like this :
std::optional<bool> getCacheEntryValue(const std::string& key, size_t hash, std::function<std::optional<bool>(FileCache*)> func) 

What's your opinion ?

XenuIsWatching:
> There are still many repetitions.
> 
> Wouldn't it be better if we refactor so we use lambdas. Something like this : std::optional getCacheEntryValue(const std::string& key, size_t hash, std::function<std::optional(FileCache*)> func)
> 
> What's your opinion ?

the problem with std::function as it heap allocates every call...there can be some overhead... maybe a template function would be better

fabricecaruso:
Fine to me. Merged

PR #2116, opened 2026-04-16
[Performance] Non-blocking media preload and VLC teardown

XenuIsWatching:
Two sources of UI stall when switching systems or opening the grid gamelist, both addressed here:

- **FileSystem preload**: `preloadFileSystemCache` previously scanned fixed subdirectories (`/images`, `/videos`, `/manuals`) under the system root. This switches it to scan the actual media paths derived from game metadata which matters most for network-backed libraries (NAS/SMB) where `stat64` latency is significant. `FolderData::getMediaDirectories()` collects these paths in one place. The overload is updated to take `std::set<std::string>` and owns the background thread and dedup logic. Preload is now triggered from `SystemView::onCursorChanged` (before the grid opens), `ViewController::preload()` (PreloadUI=true path), and `GridGameListView::populateList` (fallback for L2/R2 quick-select which bypasses `onCursorChanged`).

- **VLC teardown**: `libvlc_media_parse_stop` and `libvlc_media_release` can block the UI thread while VLC's internal parse thread acknowledges cancellation. Both are now moved off-thread in `stopVideo()`.

Before (practically unusable when scrolling)

https://github.com/user-attachments/assets/77b323da-1121-44c8-a369-fd9f4c1c41c6

After

https://github.com/user-attachments/assets/9bc57270-62f7-4fb4-885d-e2bfc9533dd9

### DIFF
--- a/es-app/src/views/gamelist/GridGameListView.cpp
+++ b/es-app/src/views/gamelist/GridGameListView.cpp
@@ -193,6 +193,8 @@ void GridGameListView::populateList(const std::vector<FileData*>& files)
 		for (auto file : files)
 			mGrid.add(formatter.getDisplayName(file, file->getType() == FOLDER && Utils::FileSystem::exists(getImagePath(file))), getImagePath(file), file);
 
+		Utils::FileSystem::preloadFileSystemCache(mRoot->getSystem()->getRootFolder()->getMediaDirectories());
+
 		// if we have the ".." PLACEHOLDER, then select the first game instead of the placeholder
 		if (showParentFolder && mCursorStack.size() && mGrid.size() > 1 && mGrid.getCursorIndex() == 0)
 			mGrid.setCursorIndex(1);
--- a/es-core/src/utils/FileSystemUtil.cpp
+++ b/es-core/src/utils/FileSystemUtil.cpp
@@ -1755,46 +1755,35 @@ namespace Utils
 			}
 		}
 #endif
-		void preloadFileSystemCache(const std::string& path, bool trySaveStates)
+		void preloadFileSystemCache(std::set<std::string> dirs)
 		{
-			if (path.empty())
+			if (dirs.empty() || !Settings::UseFileCache())
 				return;
 
-			if (!Settings::UseFileCache())
-				return;
-
+			static std::mutex preloadMutex;
 			static std::set<std::string> preloaded;
 
-			if (preloaded.find(path) != preloaded.cend())
+			std::set<std::string> todo;
+			{
+				std::lock_guard<std::mutex> lock(preloadMutex);
+				for (auto& d : dirs)
+				{
+					if (preloaded.find(d) == preloaded.cend())
+					{
+						preloaded.insert(d);
+						todo.insert(d);
+					}
+				}
+			}
+
+			if (todo.empty())
 				return;
 
-			preloaded.insert(path);
-
-			auto doWork = [&](std::string* dir)
+			std::thread([dirs = std::move(todo)]()
 			{
-				if (trySaveStates)
-				{
-					std::string systemName = Utils::FileSystem::getFileName(*dir);
-					Utils::FileSystem::getDirContent(Utils::FileSystem::combine(Paths::getSavesPath(), systemName), true);
-				}
-
-				Utils::FileSystem::getDirContent(Utils::FileSystem::combine(*dir, "/images"));
-				Utils::FileSystem::getDirContent(Utils::FileSystem::combine(*dir, "/videos"));
-				Utils::FileSystem::getDirContent(Utils::FileSystem::combine(*dir, "/manuals"));
-
-				delete dir;
-			};
-
-			std::thread thread(doWork, new std::string(path));
-
-#if WIN32		
-			if (Utils::Platform::isWindows10())
-				::SetThreadDescription(thread.native_handle(), L"PreloadFileSystemCache");
-
-			::SetThreadPriority(thread.native_handle(), THREAD_PRIORITY_LOWEST);
-#endif
-
-			thread.detach();
+				for (const auto& d : dirs)
+					Utils::FileSystem::getDirContent(d, false, true);
+			}).detach();
 		}
 
 	} // FileSystem::
--- a/es-app/src/FileData.cpp
+++ b/es-app/src/FileData.cpp
@@ -1778,6 +1778,21 @@ FolderData::~FolderData()
 	clear();
 }
 
+std::set<std::string> FolderData::getMediaDirectories() const
+{
+	std::set<std::string> dirs;
+	for (auto file : mChildren)
+	{
+		std::string img = file->getMetadata(MetaDataId::Image);
+		if (!img.empty()) dirs.insert(Utils::FileSystem::getParent(img));
+		std::string mq = file->getMetadata(MetaDataId::Marquee);
+		if (!mq.empty()) dirs.insert(Utils::FileSystem::getParent(mq));
+		std::string vid = file->getMetadata(MetaDataId::Video);
+		if (!vid.empty()) dirs.insert(Utils::FileSystem::getParent(vid));
+	}
+	return dirs;
+}
+
 void FolderData::clear() {
 	if (mOwnsChildrens)
 		for (auto* child : mChildren)
--- a/es-app/src/FileData.h
+++ b/es-app/src/FileData.h
@@ -244,6 +244,7 @@ public:
 	FileData* FindByPath(const std::string& path);
 
 	inline const std::vector<FileData*>& getChildren() const { return mChildren; }
+	std::set<std::string> getMediaDirectories() const;
 	const std::vector<FileData*> getChildrenListToDisplay();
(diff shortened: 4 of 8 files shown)


es-app/src/views/gamelist/GridGameListView.cpp
```diff
@@ -193,6 +194,36 @@ void GridGameListView::populateList(const std::vector<FileData*>& files)
 		for (auto file : files)
 			mGrid.add(formatter.getDisplayName(file, file->getType() == FOLDER && Utils::FileSystem::exists(getImagePath(file))), getImagePath(file), file);
 
+		// Pre-warm the file-existence cache in a background thread so that
```
fabricecaruso: I don't see the interest of this piece of code, since it's already done with FileSystem::preloadFileSystemCache.
Unless the medias are stored in other paths ( not in images/videos/manuals ) , there is no interest. 
+ It's far slower to do many calls to FileSystem::exists than cache files using Utils::FileSystem::getDirContent

es-core/src/components/ImageGridComponent.h
```diff
@@ -1288,10 +1388,51 @@ void ImageGridComponent<T>::loadTile(std::shared_ptr<GridTileComponent> tile, ty
...
+				tile->setImage(mDefaultGameTexture, mDefaultGameTexture == ":/cartridge.svg");
+			if (tile->hasMarquee()) tile->setMarquee("");
+			tile->setVideo("", 0);
+			tile->setFavorite(favorite);
+			tile->setCheevos(cheevos);
+
+			// Queue paths for background stat (high-priority front insert so
+			// newly-visible tiles are processed before pre-warm tail work)
+			{
+				std::lock_guard<std::mutex> lock(mTileStatMutex);
```
fabricecaruso: The code is very complex, just to defer cases where some media is not in the file cache....
-> If the gamelist is "normal", the FileSystem::preloadFileSystemCache list every medias in the cache, so I don't see a case where all this is useful.

fabricecaruso (commented):
I'm sorry for my comments, but, to me, this is useless.
Give some precise cases where it's useful, please.

XenuIsWatching:
these file IO are rather blocking operations, and if it's on a NAS these can block the main thread (where the UI runs) it can block for a while making it appear as if the screen is frozen which is extremely noticeable when scrolling. 

Put them on a separate thread, makes it so they are not blocking anymore

i can give a video of the before and after later here

XenuIsWatching:
> these file IO are rather blocking operations, and if it's on a NAS these can block the main thread (where the UI runs) it can block for a while making it appear as if the screen is frozen which is extremely noticeable when scrolling.
> 
> Put them on a separate thread, makes it so they are not blocking anymore
> 
> i can give a video of the before and after later here

added in to the summary post

es-app/src/views/gamelist/GridGameListView.cpp
XenuIsWatching: Thanks for your review!

preloadFileSystemCache is guarded by #if WIN32 and is a no-op on Linux

I've updated the pre-warm to call getDirContent. I can actually see the difference and images stream in noticeably faster with this approach, even though the render thread was never blocked either way

After the change (compare to the after rin the summary)
https://github.com/user-attachments/assets/9bc57270-62f7-4fb4-885d-e2bfc9533dd9

fabricecaruso:
As you said, it's true, preloadFileSystemCache is guarded by #if WIN32 

You are doing your tests on linux ? 
If so, can you tell me if there is a performance difference if you enable preloadFileSystemCache for linux, compared to the code you wrote ?

I haven't tested this under Linux (because I use Linux only in VirtualBox VMs), but I can say that under Windows, preloadFileSystemCache is very effective.

Wouldn't it be enough to use preloadFileSystemCache on linux too - and just remove the #if condition in preloadFileSystemCache - instead of modifying the grid ?

Changes you propose on the grid are big and make the grid more complex ( it's already complex ! )

XenuIsWatching:
> As you said, it's true, preloadFileSystemCache is guarded by #if WIN32
> 
> You are doing your tests on linux ? If so, can you tell me if there is a performance difference if you enable preloadFileSystemCache for linux, compared to the code you wrote ?

yes, and it's the same as before any changes, i tried re-running before these changes with those #if WIN32 guards removed and it's just as stuttery

> I haven't tested this under Linux (because I use Linux only in VirtualBox VMs), but I can say that under Windows, preloadFileSystemCache is very effective.

What difference does the OS make here? I'm going across a NAS filesystem here, so it's going across a network for File IO. Latencies are gonna be HUGE. What is the disk type for where you ran your tests on?

> Wouldn't it be enough to use preloadFileSystemCache on linux too - and just remove the #if condition in preloadFileSystemCache - instead of modifying the grid ?
> 
> Changes you propose on the grid are big and make the grid more complex ( it's already complex ! )

`preloadFileSystemCache` fires in `onCursorChanged` at the same moment the grid starts rendering tiles... they race. On a NAS over Samba, the preload background thread can't win that race: the render thread hits uncached paths and blocks on `stat64` before the preload finishes, which is exactly what causes the stutter.
The grid change fixes this at a different level... `loadTile()` itself becomes non-blocking. Instead of calling `stat64` on the render thread for an uncached path, it shows a placeholder immediately and defers the stat work. The render thread never waits on I/O, regardless of whether the preload has finished. That's why enabling `preloadFileSystemCache` here produces no improvement, while the grid change does.

es-core/src/utils/FileSystemUtil.cpp
```diff
@@ -204,6 +204,22 @@ namespace Utils
 				return std::nullopt;
 			}
 
+			// Non-blocking probe: returns true if the key's existence is already
+			// recorded in the cache (either exists or known-not-to-exist).
+			// Never calls stat64 — safe to call from the render thread.
+			static bool isCached(const std::string& key)
```
fabricecaruso, line 226: This will conflict with the other PR I merged.

es-app/src/views/gamelist/GridGameListView.cpp
fabricecaruso: Still not convinced by this peace of code that does almost the same thing as preloadFileSystemCache, but does it when the grid load ( preloadFileSystemCache is when the System is activated in the SystemView ).
I really think this should be moved to preloadFileSystemCache.

es-core/src/utils/FileSystemUtil.cpp
XenuIsWatching: yeah, that was a function added in that other pr as well...

i removed it

fabricecaruso:
I'm truly sorry, but I can't convince myself this PR is a good idea.

What you did is a quite complex code that only handles the basic case - which is kept here for retrocompatilibity only and which I'd like to get rid of one day eventually ) -> Grids without ItemTemplates ! 

Take the example of the default display of the Carbon theme's grid (icon view -> https://github.com/fabricecaruso/es-theme-carbon/blob/master/views/icons.xml); it's now based on the ItemTemplate description. The same goes for the "Content" view.

The code in your PR doesn't handle cases where there is an ItemTemplate. If there is a template, you never reach mTileStatQueue.push_front(..), and you never cache anything except with the call to preloadFileSystemCache.

Also, now with bindings, lot of themes have now complex conditions like `<path>firstfile({game:fanart},{game:image},"${themePath}/_systemmedia/_fanartsysteme/fresh/${system.theme}.jpg")</path>`.

How to handle this ? That would be a complex piece of code to write, and it's not certain that it's worth it.
Anyway your code won't improve anything about bindings and ItemTemplates.

I still think modifying preloadFileSystemCache, removing the #if WIN32 and adapting the method in order it applies to linux too, is the only halfway acceptable solution.

XenuIsWatching:
> I'm truly sorry, but I can't convince myself this PR is a good idea.
> 
> What you did is a quite complex code that only handles the basic case - which is kept here for retrocompatilibity only and which I'd like to get rid of one day eventually ) -> Grids without ItemTemplates !
> 
> Take the example of the default display of the Carbon theme's grid (icon view -> https://github.com/fabricecaruso/es-theme-carbon/blob/master/views/icons.xml); it's now based on the ItemTemplate description. The same goes for the "Content" view.
> 
> The code in your PR doesn't handle cases where there is an ItemTemplate. If there is a template, you never reach mTileStatQueue.push_front(..), and you never cache anything except with the call to preloadFileSystemCache.
> 
> Also, now with bindings, lot of themes have now complex conditions like `<path>firstfile({game:fanart},{game:image},"${themePath}/_systemmedia/_fanartsysteme/fresh/${system.theme}.jpg")</path>`.
> 
> How to handle this ? That would be a complex piece of code to write, and it's not certain that it's worth it. Anyway your code won't improve anything about bindings and ItemTemplates.
> 
> I still think modifying preloadFileSystemCache, removing the #if WIN32 and adapting the method in order it applies to linux too, is the only halfway acceptable solution.

Alright, all the `ImageGridComponent` code has been removed entirely

XenuIsWatching:
apparently, in one push i made, i rather corrupted this PR earlier, as I did a lot of local 'moving things' around, i've repushed it then rebased it up and fixed the conflict

@fabricecaruso please take a look 😄

PR #2124, opened 2026-05-08
feat: notify user when a manual update file is available

zognic:
## Summary

When a user manually downloads a Batocera update and places `boot.tar.xz` in `/userdata/system/upgrade/`, EmulationStation had no way to detect it or prompt the user to apply it.

- **Auto-detection at startup** — a new `CheckManualUpdateComponent` watcher fires 5 seconds after ES starts, then checks every minute. When `boot.tar.xz` is found, a `GuiMsgBox` popup appears asking the user to apply the update immediately or later.
- **Menu entry** — a conditional "APPLY MANUAL UPDATE" entry is added to *Settings > Updates* whenever the file is present, allowing the user to trigger it manually at any time.
- **Reuses existing update flow** — `ThreadedUpdater` and `AsyncNotificationComponent` are reused for progress display; `batocera-upgrade manual` is called instead of `batocera-upgrade`.

## Changed files

| File | Change |
|------|--------|
| `ApiSystem.h/.cpp` | `hasManualUpdate()` checks for `boot.tar.xz`; `updateSystem()` gets a `manual` bool flag |
| `NetworkThread.h/.cpp` | New `CheckManualUpdateComponent` watcher + `OnWatcherChanged` handler showing the popup |
| `guis/GuiUpdate.h/.cpp` | `GuiUpdate::startUpdate(window, manual)` static method; `ThreadedUpdater` supports manual mode |
| `guis/GuiMenu.cpp` | Conditional "APPLY MANUAL UPDATE" menu entry |

## How to test

1. SSH into Batocera and create the trigger file:
   ```bash
   touch /userdata/system/upgrade/boot.tar.xz
   ```
2. Restart EmulationStation — a popup should appear after ~5 seconds.
3. Click **LATER** — the popup dismisses and the entry appears in *Settings > Updates*.
4. With a real `boot.tar.xz`, clicking **YES** launches `batocera-upgrade manual` with a progress notification.


zognic:
screenshot-2026 05 08-22h02 02

screenshot-2026 05 08-22h22 56

nadenislamarre:
i don't get the aim.
who updates like it ?
i understand from the pr that the file is checked every minute, but according to your scenario, it is just at boot only once.
if that's for offline batocera system, wouldn't it make more sense to check if a boot.tar.xz file exists at the root of the disk and use it to upgrade manually ?

zognic:
Thanks for the feedback — you're right on all points.

**Use case:** offline/air-gapped systems, or users who prefer to download updates manually and apply them via USB. The `batocera-upgrade manual` command already existed for this, but had no UI to trigger it from ES.

**Reworked approach (just pushed):**

Instead of checking a fixed internal path, `findManualUpdateFile()` now scans `/media/*/boot.tar.xz` — i.e. the root of any mounted external device (USB, SD card). It falls back to `/userdata/system/upgrade/boot.tar.xz` for users who placed the file there manually (e.g. via SCP).

Workflow:
1. User downloads `boot.tar.xz` and places it at the **root of a USB drive**
2. Plugs in the drive — ES detects it within 1 minute and shows a popup
3. On confirmation, the file is **copied to `/userdata/system/upgrade/`** (with a "COPYING..." notification), then `batocera-upgrade manual` runs with the existing progress display

The watcher now also resets when the file disappears, so re-plugging a different USB works correctly.

nadenislamarre:
es is a frontend.... most of things here should not be done here. and the usb plug in is an event, not something to check every minute on all devices. the code is maybe correct, but the way it is integrated is wrong and doesnt respect code responsabilities.

crcerror:
I think a manual update feature from ES is edge use but can maybe handy.
We have a list `<STABLE> <BUTTERFLY>` and maybe can add `<MANUAL>`
From there do the file checks und use the ES badge for progress **BUT ATTENTION ....**

A manual upgrade can be fatal if you flash the wrong architecture on your current machine ... so rather edge use and not wanted.

**EDIT**

To avoid a wrong flashed image you can try `LC_ALL=C grep -aRl -F x86_64 boot.tar.xz`
I haven't tested on all platforms but the search string can be found in several seconds with his method
@dmanlfc maybe useful for batocera-upgrade, too? I mean you can send a warning during extraction so user can break the process or do you think to error-prone? So only reliable way is to read `batocera.board`

PS: tried `batcoera-xtract` to extract that single file beforehand .... yes works but is slow :(

fabricecaruso:
Update process are managed by external scripts.
Such a thing should be 100% managed by scripts ( "batocera-config canupdate" for detection, "batocera-upgrade" for the installation process )
I don't see the need for the CheckManualUpdateComponent and findManualUpdateFile thing : "batocera-config canupdate" should do the job instead, asking the script if there is some update available.
In the script's code : If a local update file exists, then the priority should is the file, and no need for an online check.
Also, There is absolutely no interest to add a new menu option to tell it's upgrading from a local file : "batocera-config canupdate" can return a custom version string do when you launch the update, it can tell it's a custom file : just return the version string, followed be some "(custom file)" information, and that's it. Then when you run "batocera-upgrade", as the priority is set to the local file, there is no need for a different menu entry.

In my opinion, all this has nothing to do inside ES code. The PR should be abandonned, and managed in batocera's scripts.

crcerror:
Ah priority task, driven inside the script is easy to manage. For myself - I will not work on this

PR #2140, opened 2026-05-20
http api : addgames/removegames : include folders in fileMap

aderumier:
When a game with a subdirectory path (e.g. ./subdir/game.zip) was added via the HTTP API, the fileMap was built with only GAME nodes. findOrCreateFile could not find the existing subfolder causing the game to appear in the root folder instead of the correct subdirectory.


Nelfe80:
Do you see a connection to this issue?
https://github.com/batocera-linux/batocera-emulationstation/issues/2141
Could you add the bug fix to your PR?

PR #2144, opened 2026-05-23
Add option to enable/disable video previews

spycat88:


fabricecaruso:
I don't agree with this PR and personnally will refuse to merge it because it's clearly not a good idea ! 

This option is useless and could cause problems with many themes that use videos (in the systemview notably).

Most themes already have an option to disable videos previews.
If the theme you use don't have such an option : contact the theme designer instead.

spycat88:
The PR was submitted as a courtesy because this functionality has already been used in ROCKNIX and JELOS for years without causing the issues you're describing.

Calling it “clearly not a good idea” while dismissing existing real-world usage doesn’t feel like an objective review. If there are concrete technical concerns specific to this implementation, I’m happy to discuss them, but rejecting it outright based on assumptions about themes isn’t particularly constructive.

Themes already handling video toggles is also not really an argument against having a core option — especially for users on themes that don’t expose one consistently.

PR #2154, opened 2026-06-10
Fix windows build (updatesystem override missing bool fromlocalmedia)

Tartifless:
Caused by Commit 3290509


Tartifless:
@fabricecaruso , this fixes Windows build following commit 3290509

PR #2155, opened 2026-06-10
WIN32-Add FPS limiter

Tartifless:
This adds a setting in dev options to have a FPS limiter with predefined options: 30, 60, 90, 120, NONE


Tartifless:
@fabricecaruso for review

PR #2156, opened 2026-06-10
Fix Bezel (decorations menu)

Tartifless:
Image is loaded with setImage(imageUrl, false, maxSize), when positionning cursor, ImageComponent not yet loaded have a reduced height, when textures finally arrive, it creates a resizing but mSelectorBarOffset is never recalculated.

The fix here is to create all lines with maxsize, there might be a better fix (dynamic recalculation) but i don't want to touch a component used in all menus.


Tartifless:
@fabricecaruso : small fix for the decorations menu

PR #2160, opened 2026-06-15
Add LED support for R36 Ultra

ImanolBarba:
Title says all.

As explained in the commits, user can change the LED colours and power LED on/off via the settings menu.

Should the user want the pulse or rainbow effects, they should change the leds.conf from the batocera-led-handheld script

### DIFF
--- a/es-app/src/ApiSystem.cpp
+++ b/es-app/src/ApiSystem.cpp
@@ -716,6 +716,33 @@ bool ApiSystem::setPowerLedGameForce(std::string selected)
 	return executeScript("batocera-gameforce powerLed " + selected);
 }
 
+bool ApiSystem::setButtonColorR36Ultra(const std::string& selected)
+{
+	std::map<std::string, std::string> r36UltraModeMap = {
+		{"off", "0 0 0"},
+		{"red", "255 0 0"},
+		{"yellow", "255 255 0"},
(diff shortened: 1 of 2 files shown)


dmanlfc:
see the correct areas to do this in `bool ApiSystem::getLED(int& red, int& green, int& blue)` onwards.
this makes it consistent across the device variants we support.
this will allow the users to choose their colours and brightness accordingly.

ImanolBarba:
The problem is, they can't.

The Ultra does not support setting brightness or colours, it has 9 static modes that are controlled by an IC, you can only toggle one of the 9 modes or off

My original plan was to enable the sliders and approximate whatever colour that was to one of the nine modes but I think the user experience is worse that way

dmanlfc:
Yeah it's similar to the Mangmi Air X. You just do a for loop to write the same colour to the nine nodes.

ImanolBarba:
Right, but the modes can't be changed or reprogrammed, they're baked into a separate microcontroller we can't control, we can only request one of the modes.

ImanolBarba:
see: https://github.com/ImanolBarba/keros-leds

dmanlfc:
Ok understood, that's a shame.

ImanolBarba:
Ugh, I forgot to include the header on the commit, sending another PR, apologies!

PR #2167, opened 2026-06-30
Add parsing for UTF-8/UTF-16 ID3 tags in MP3

lbrpdx:
screenshot-2026 06 30-13h29 32

Updated to the latest id3v2 library, with git external dependency from now on. Also implemented a parser to properly decode UTF-8 and UTF-16 ID3 tags. Seems to be working well, but of course needs testing with as many MP3s as possible.

@fabricecaruso can you please check it is OK with Windows compilation? On Linux, it needs to install the additional dynamic library libid3v2lib.so I assume something equivalent needs to be done under Windows as well.


fabricecaruso:
This library does not compile under windows, and it's not possible to fix as it's an external. The source code uses unsupported syntaxes.
Either, we have to revert this PR
Either we should not use it as an external but an "internal source" and fix everything manually ( which is complex )
Either we have to exclude all this on Windows.

fabricecaruso:
Disabled for Windows https://github.com/batocera-linux/batocera-emulationstation/pull/2175

PR #2170, opened 2026-07-07
Fix frontend frame pacing when VSync does not throttle

RamazanKara:
## Summary

- add automatic frontend frame pacing when VSync is enabled but no explicit FPS limit is configured
- base the automatic limit on the active SDL display refresh rate, with the existing Windows display-frequency fallback and a 60 FPS final fallback
- run frame limiting after buffer swap so working VSync is not double-throttled

## Why

On some virtual-display streaming setups, such as Apollo/Moonlight, the GL swap interval can fail to provide practical throttling even when VSync is enabled. The frontend then renders as fast as possible, which can cause visible flicker unless users configure an external frame limiter or manually set an FPS limit.

This keeps explicit FPS limit settings authoritative, but makes the default VSync path behave like users expect: if swap does not block, the frontend still paces itself to the display refresh rate.

## Validation

- `git diff --check HEAD~1..HEAD`
- `cmake -S . -B build` was attempted, but this WSL environment is missing the FreeImage development package and `sudo` requires a password, so the full configure/build could not be completed here.

### DIFF
--- a/es-app/src/main.cpp
+++ b/es-app/src/main.cpp
@@ -20,6 +20,7 @@
 #include <SDL_events.h>
 #include <SDL_main.h>
 #include <SDL_timer.h>
+#include <SDL_video.h>
 #include <iostream>
 #include <time.h>
 #include "LocaleES.h"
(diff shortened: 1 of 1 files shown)


RamazanKara:
Replaced by #2171 from a neutral branch name.

PR #2174, opened 2026-07-14
Fix to enable filtering based on year

TrickoUK:
Fix was found with Claude CLI, but seems like filtering based on year has been broken since a change in 2021. 
Previously you get an empty picker list when selecting the year category, now you get the real list.
Tested by building it into batocera.

IMG_3618
IMG_3619

### DIFF
--- a/es-app/src/FileFilterIndex.cpp
+++ b/es-app/src/FileFilterIndex.cpp
@@ -362,9 +362,11 @@ std::string FileFilterIndex::getIndexableKey(FileData* game, FilterIndexType typ
 		return Utils::String::toInteger(game->getMetadata(MetaDataId::PlayCount)) == 0 ? "FALSE" : "TRUE";		
 
 	case YEAR_FILTER:
+	{
 		key = game->getMetadata(MetaDataId::ReleaseDate);
 		key = (key.length() >= 4 && key[0] >= '1' && key[0] <= '2') ? key.substr(0, 4) : "";
-		return key;
+		break;
+	}
 
 	case CHEEVOS_FILTER:
 	{


fabricecaruso:
The current code works fine with my gamelists. 
Can you explain me what changes with a break instead of a return ret, as the final return will be the same value ?
Are you really sure you fixes does something ????

TrickoUK:
Sorry I should have been clearer, its specifically when there are games within a list which dont have a year value set that will trigger the behaviour. Arcade systems like FBneo have more chance to show that up I think, certainty for me.

In terms of the details in the change, I'll copy the whole description from Claude here for clarity rather -


One file, one `switch` case: `es-app/src/FileFilterIndex.cpp`,
`FileFilterIndex::getIndexableKey()`, the `YEAR_FILTER` case (currently
uncommitted in this working tree — `git diff` shows it, nothing pushed yet):

```diff
 	case YEAR_FILTER:
+	{
 		key = game->getMetadata(MetaDataId::ReleaseDate);
 		key = (key.length() >= 4 && key[0] >= '1' && key[0] <= '2') ? key.substr(0, 4) : "";
-		return key;
+		break;
+	}
```

**Root cause.** Every filter type in `getIndexableKey()` (publisher, genre,
family, ratings, ...) computes its `key` and falls through to a shared tail at
the end of the function:

```cpp
if (key.empty() || (type == RATINGS_FILTER && key == "0 STARS"))
    return UNKNOWN_LABEL;
return Utils::String::toUpper(key);
```

That tail is what turns "this game has no value for this attribute" into the
sentinel `"UNKNOWN"` — which `manageIndexEntry()` (same file, ~line 1069)
then deliberately excludes from ever showing up as a selectable option
(`INCLUDE_UNKNOWN` is `false`). `YEAR_FILTER` used to reach that tail via
`break;` like everything else. Commit `6b3fe9263` ("Misc Optimisations &
fixes", 2021-02-10) mechanically swept several cases over to `return ...;`
directly — correct for cases that always return a non-empty literal
(`PLAYER_FILTER`'s `"TRUE"`/`"FALSE"`, etc.), but `YEAR_FILTER` got caught in
the same sweep despite being able to legitimately produce an **empty** key
(any game with no/invalid release date). From that commit on, an empty year
key was returned raw instead of being normalized to `"UNKNOWN"`, so
`manageIndexEntry()`'s "skip unknown" guard never fired for it — every game
lacking a recognized release date silently inserted a spurious `""` entry
into `yearIndexAllKeys`, corrupting the map that the filter menu
(`GuiGamelistFilter::addFiltersToMenu()`) turns into the popup you see when
you select "YEAR". Restoring `break;` makes Year behave exactly like every
other filter: valid years still extract identically (the
`key[0] >= '1' && key[0] <= '2'` heuristic was already correct — verified
against real gamelist data, see §3), and games with no date now correctly
disappear into `UNKNOWN_LABEL` instead of polluting the picker.

No other file needed to change — `manageYearEntryInIndex()`, `addToIndex()`,
`SystemData::indexAllGameFilters()`, and `GuiGamelistFilter` are all
filter-type-agnostic and already correct; they were just being fed a broken
map for Year specifically.

PR #2177, opened 2026-07-17
Fix update of gamelist when changing multidisk location (m3u)

Tartifless:


Tartifless:
@lbrpdx  @dmanlfc : for review

es-app/src/SystemData.cpp
```diff
@@ -197,20 +197,27 @@ void SystemData::removeMultiDiskContent(std::unordered_map<std::string, FileData
 		{
 			if (it->getType() == GAME && it->hasContentFiles())
 			{
+				std::string stamp = "mtime:" + std::to_string(
+					(long long)Utils::FileSystem::getFileModificationDate(it->getPath()).getTime());
```
fabricecaruso, line 201: This bothers me because it will hurt performance, since it will inevitably involve access to the file system as the File Modification Date is not cached.

Tartifless: Not on windows, maybe on linux.

FindFirstFileEx already returns ftLastWriteTime.
Mtime is already in memory at boot.

PR #2182, opened 2026-07-27
Declare Per-Monitor-V2 DPI awareness at startup on Windows

sdornan:
## Summary
- On Windows, ES has never declared DPI awareness (no manifest, no `SetProcessDpiAwareness*`/`SetProcessDPIAware` call anywhere in the codebase). On displays scaled above 100% (e.g. 4K @ 125%/150%), Windows therefore treats the process as DPI-unaware: it hands ES a virtualized logical resolution and bitmap-stretches the rendered output back up, causing blur and fullscreen/window-sizing issues. Today the only workaround is manually enabling "Override high DPI scaling" in the exe's Compatibility settings.
- `SDL_WINDOW_ALLOW_HIGHDPI` is already set on the window (`Renderer.cpp`), but it has no effect without process-level DPI awareness declared first.
- This declares the process Per-Monitor-V2 DPI aware as the first thing in `main()`, before SDL/window init, so the manual override is no longer needed.
- Resolves `SetProcessDpiAwarenessContext` → `SetProcessDpiAwareness` → `SetProcessDPIAware` dynamically via `GetProcAddress`, falling back through progressively older APIs, so the same binary keeps working on older Windows versions where the newer entry points don't exist, without needing an OS-version-gated manifest.
- Entire change is wrapped in the existing `#ifdef WIN32` block (same one guarding `#include <Windows.h>`); no effect on Linux/other platform builds.

## Test plan
- [ ] On Windows with display scaling set above 100% (e.g. 4K @ 150%), remove any manual DPI compatibility override on the exe, launch ES, and confirm it renders fullscreen at native resolution without blur.
- [ ] Confirm behavior is unchanged at 100% scaling.
- [ ] Confirm Linux build is unaffected (change is fully inside `#ifdef WIN32`).

### DIFF
--- a/es-app/src/main.cpp
+++ b/es-app/src/main.cpp
@@ -49,64 +49,6 @@
 #include <Windows.h>
 #include <direct.h>
 #define PATH_MAX MAX_PATH
-
-// Declared locally (instead of pulled from ShellScalingApi.h / a newer
-// winuser.h) so this builds regardless of which Windows SDK version ES is
-// compiled against; on older SDKs the constant/type simply doesn't exist yet.
-#ifndef DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2
-DECLARE_HANDLE(DPI_AWARENESS_CONTEXT);
-#define DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2 ((DPI_AWARENESS_CONTEXT)-4)
-#endif
-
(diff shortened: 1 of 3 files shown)


fabricecaruso:
Please move all the system specific code into platform.cpp

sdornan:
Done — moved it into `Utils::Platform::setDpiAwareness()` in `es-core/src/utils/Platform.h`/`.cpp`, next to `isWindows10()`/`isWindows11()`. `main.cpp` now just calls it.

PR #2185, opened 2026-08-06
Fixing a typo in UI: ccp and po files

NotaInutilis:
Hope this is alright! I followed the translation guide with VSCodium as a find and replace tool.

The fun thing about this typo is that every locale has corrected it already!


nadenislamarre:
i merged, but i reverted all.
cause "..." as a single char is utf8 and is not allowed for translations, making the application not to compile anymore.
2185 and 2189 have been reverted. (ooops i did a bad push -f on master cause too many commit and merges to revert.)

NotaInutilis:
Ah yeah sorry, I'm not knowledgeable enough to actually test compilation!

Unicode ellipsis is already in some translations, so it's alright in .po files but it should be kept out of the .cpp files, right?

(And I actually did this using several commits so they could be easily cherry picked! Oops indeed!)

PR #2186, opened 2026-08-07
Windows: enumerate video modes on the selected monitor

Nelfe80:
I hit this while testing a CRT as a secondary screen on Windows.

The CRT has a custom 640x400 @ 70 Hz mode made with CRU. Windows and RetroArch can use it, but ES does not show it in Advanced Video Resolution. If I make the CRT the primary screen, the 70 Hz mode appears immediately.

Looking at the Windows code, getVideoModes(output) receives an output but EnumDisplaySettings(nullptr, ...) always enumerates the default display.

This PR makes the Win32 code ask Windows for the modes of the correct screen:

use the explicit output when one is provided
otherwise resolve ES MonitorID from the SDL display to the Windows \\.\DISPLAYx device
keep the existing mode list and refresh-rate handling unchanged

There is no hardcoded 70 Hz list here: refresh rates still come directly from dmDisplayFrequency. The problem is mainly that ES was asking Windows for the modes of the wrong screen.

My setup that exposed it:

Windows 11
RTX 3070 + Intel UHD 770
CRT on a secondary output
custom resolutions with CRU
640x400 @ 70 Hz

Before: 70 Hz only appears in ES when the CRT is set as the primary screen.

Expected with this change: ES can enumerate the modes of the selected secondary monitor without having to change the Windows primary display.

I kept the change intentionally limited to the Win32 video-mode enumeration path.

### DIFF
--- a/es-app/src/Win32ApiSystem.cpp
+++ b/es-app/src/Win32ApiSystem.cpp
@@ -889,7 +889,7 @@ void Win32ApiSystem::updateEmulatorLauncher(const std::function<void(const std::
 bool Win32ApiSystem::canUpdate(std::vector<std::string>& output)
 {
 	// Update using 'es-checkversion.cmd' scripts ?
-	std::string esUpdateScript = getScriptPath("es-checkversion");
+	std::string esUpdateScript = getScriptPath("es-update");
 	if (!esUpdateScript.empty())
 	{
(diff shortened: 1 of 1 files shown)


Tartifless:
@nadenislamarre @lbrpdx @dmanlfc , can you merge this one, it's win32api only

PR #2189, opened 2026-08-11
Update FR locale

NotaInutilis:
Translating new lines + multiple fixes and general editing

### DIFF
--- a/locale/lang/fr/LC_MESSAGES/emulationstation2.po
+++ b/locale/lang/fr/LC_MESSAGES/emulationstation2.po
@@ -104,7 +104,7 @@ msgid "Loading gamelists"
 msgstr "Chargement des listes de jeux"
 
 msgid "Loading..."
-msgstr "Chargement..."
+msgstr "Chargement…"
 
 msgid "MENU"
 msgstr "MENU"
@@ -676,7 +676,7 @@ msgid "CONTENT DOWNLOADER"
 msgstr "TÉLÉCHARGEMENT DE CONTENU"
 
 msgid "SELECT CONTENT TO INSTALL/REMOVE"
-msgstr "SÉLECTIONNER LE CONTENU À INSTALLER / SUPPRIMER"
+msgstr "SÉLECTIONNER LE CONTENU À INSTALLER/SUPPRIMER"
 
 msgid "UPDATE INSTALLED CONTENT"
 msgstr "MÀJ LE CONTENU INSTALLÉ"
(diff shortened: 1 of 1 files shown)


nadenislamarre:
thanks a lot for all these fixes.
most of translations are ok.
some are discutable : the one i noted bellow (lbrpdx, modhack, what do you think about them ? i merge cause 99% is better than the existing translations, 1% is always suggest to discussion on translations)

SUCCÈS RÉTRO => RETROACHIEVEMENTS
"INDEXATION DES JEUX" => INDEX DES JEUX (but maybe the english wording is not the best)
RAZ =>  "RÉINITIALISER" (this one was probably for size issues)

msgid "REALLY UPDATE?"
-msgstr "VOULEZ-VOUS VRAIMENT METTRE À JOUR ?"
+msgstr "CONFIRMER LA MISE À JOUR ?"

msgid "START GAME"
-msgstr "DÉMARRER LE JEU"
+msgstr "DÉMARRER LA PARTIE"

msgid "LEFT STICK PRESS"
-msgstr "PRESSION MANETTE GAUCHE"
+msgstr "CLIC STICK GAUCHE"

msgctxt "joystick"
msgid "LEFT SHOULDER"
-msgstr "L1"
+msgstr "TRANCHE GAUCHE (L1)" (strange naming)

zognic:
` TRANCHE GAUCHE (L1) ` is a Sony Wording , Nintendo use `L1/R1 `, Microsoft use `LB/RB `(Bumber).
 L / R is better imho
 
 
 I will keep `"VOULEZ-VOUS VRAIMENT METTRE À JOUR ?"` It's a warning, not a simple confirmation.
 
 And for this one, I suggest (because it's not a Party)
 ```
 msgid "START GAME"
-msgstr "DÉMARRER LE JEU"
+msgstr "LANCER LE JEU"
 ```

NotaInutilis:
Yeah, there are some where I was unsure so I'm glad we can discuss it! (It's actually not stuff I was unsure about!)

RetroAchievements as one word is a proper noun, if it was "retro achievements" then "succès rétro" would be great.

Indexation fits better because we're not looking at the index itself in the menu, but the indexing process.

RAZ is not common so it might not be understood. If we keep it short, maybe R.À.Z. whould be more explicit?

I had to rack my memories bit to find "Bouton de tranche", I remember this is the term we used 20+ something years ago to discuss shoulder buttons on SNES or GBA hardware. The PS style parenthesis were already there in the previous edit. I'm not sure the trigger/bumper difference is really obvious in French, we tend to use gâchette for everything.

nadenislamarre:
RAZ is not common so it might not be understood. If we keep it short, maybe R.À.Z. whould be more explicit? : yes if that can help.
don't hesitate to join on discord i'm frequently here. and at least by phone. susan34
and i'm french you know :-P
english can use ascii only.

PR #2197, opened 2026-08-22
improve-watchers

Mikhailzrick:
This PR does 2 things:

1. Switches from the older SDL_GetTicks() to a more modern use of std::chrono::steady_clock. This uses monotonic time and removes the SDL dependency.

2. Pauses WatchersManager while a game is running and resumes when returning back to ES. This pauses the watcher threads instead of waking periodically to poll components when there is no active UI to update.


dmanlfc:
looks ok @fabricecaruso thoughts?

PR #2201, opened 2026-08-28
PR2186 - addendum

Tartifless:
This comes as an addition to Commit 7508bc2

This PR secures the case when someone would enable/create a getAvailableVideoOutputDevices() method for windows.
This would use display names like HDMI-1 instead of \\.\DISPLAY1, and would instantly fail in EnumDisplaySettingsA with first index, outputting an empty list (or AUTO only).

### DIFF
--- a/es-app/src/Win32ApiSystem.cpp
+++ b/es-app/src/Win32ApiSystem.cpp
@@ -998,53 +998,109 @@ static std::string getWindowsDisplayDeviceName(int monitorId)
 	return monitorInfo.szDevice;
 }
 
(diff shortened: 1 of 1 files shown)


Tartifless:
@lbrpdx : can you merge this windows addendum to Commit https://github.com/batocera-linux/batocera-emulationstation/commit/7508bc2b077f0671bd2ba998f9c6e24fd1f09ed8 ?

lbrpdx:
I trust you tested it

Tartifless:
Yes the behaviour does not change, it's only a failsafe in case the method to get videodevice is enabled in windows once

PR #2202, opened 2026-08-29
[WIN32] Fix crashes, freezes and duplicate controllers on joystick ho…

Tartifless:
Users reported EmulationStation crashing or hanging when a wireless controller powers itself off after a period of inactivity. Investigating that led to several distinct defects in the joystick add/remove path, all fixed here.

Reproduced with an 8BitDo Ultimate C (2.4G), which is handled by SDL's HIDAPI/RAWINPUT backends. Pads handled by the Windows backend (DInput/XInput) do not hit most of these, which is why it only affected some users.

**TESTS**
Tested extensively over the last week by myself and provided to users (tested with multiple XInput controllers and multiple connexions / disconnections with up to 6 controllers connected), also tested with dualsenses and switch pro controllers.

**What this fixes**
**Infinite freeze in the HIDAPI stabilisation loop**.
The attempts counter added in 2add5f70 was never incremented, so the "max 500ms" bound was never enforced. The loop only exited once SDL_NumJoysticks() returned the same value twice. A device whose count keeps oscillating — a 2.4G dongle re-enumerating, a pad sleeping and waking — hung the main thread for good. Windows shows this as "not responding" and users report it as a crash.

**Two std::string constructions from a possibly-NULL const char***.
SDL_JoystickPathForIndex returns NULL for an out-of-range index and for HIDAPI devices no longer in the list; SDL_JoystickNameForIndex returns NULL for an out-of-range index. Both were assigned straight into a std::string. SDL_JoystickName is now null-checked too.

**Stale device index on SDL_JOYDEVICEADDED**.
ev.jdevice.which is a device index there, not an instance id, and it is only valid when the event is posted. The main loop drains the whole queue at once, and a SDL_JOYDEVICEREMOVED handled just before rebuilds the entire SDL enumeration in between. The index is now validated against SDL_NumJoysticks().

**getInputConfigByDevice was inserting null entries**. 
std::map::operator[] default-constructs on a missing key, so every event carrying an already-removed instance id left a permanent {id, nullptr} in mInputConfigs. Most call sites had been patched to tolerate this; the SDL_JOYBATTERYUPDATED handler had not and dereferenced it. Uses find() now, plus a null check on the battery handler.

**Out-of-bounds write on mPrevAxisValues**.
Raw int arrays sized from SDL_JoystickNumAxes(), with the size never stored and ev.jaxis.axis never bounded. Replaced with std::vector<int> and a bounds check; also removes the manual delete[] and the leak that came with it.

**Duplicate controller after a hot reconnection**.
SDL can expose the same physical pad through two backends: XInput and DInput skip a device when RAWINPUT_IsDevicePresent() returns true, but that loses the race when XInput enumerates before RAWINPUT has registered it, and the Windows backend only re-enumerates on the next WM_DEVICECHANGE, so the duplicate sticks until restart.

Observed with the 8BitDo Ultimate C powered off and back on while ES runs — VID 2DC8 / PID 3106 shows up both as (Ultimate Wireless Controller) (GUID ...7200, driver byte 'r', RAWINPUT) and as Contrôleur XBOX 360 pour Windows (GUID ...7701, driver byte 'w', WINDOWS). It takes two player slots and writes a second inputConfig node into es_input.cfg.

A cold start keeps only the RAWINPUT view, so this reproduces that: the redundant XInput view carries no HID device path while the RAWINPUT one does, so a path-less device is dropped when another device with the same vendor and product was enumerated with a path. Two identical controllers each expose their own real path and are never merged.

**Notes**

Linux behaviour is unchanged apart from the null checks, which apply to the shared code paths.
The underlying enumeration race belongs to SDL and is worth reporting upstream; the guard here is warranted either way.

### DIFF
--- a/es-core/src/InputManager.cpp
+++ b/es-core/src/InputManager.cpp
@@ -19,6 +19,7 @@
 #include "GunManager.h"
 #include "renderers/Renderer.h"
 #include <fstream>
+#include <set>
 
 #ifdef HAVE_UDEV
 #include <libudev.h>
@@ -211,7 +212,11 @@ InputConfig* InputManager::getInputConfigByDevice(int device)
 	if(device == DEVICE_GUN)
 		return mGunInputConfig;
 	
-	return mInputConfigs[device];
+	auto it = mInputConfigs.find(device);
+	if (it == mInputConfigs.cend())
+		return nullptr;
+
+	return it->second;
 }
 
 void InputManager::clearJoysticks()
@@ -228,10 +233,6 @@ void InputManager::clearJoysticks()
 			delete iter->second;
 
 	mInputConfigs.clear();
-
-	for (auto iter = mPrevAxisValues.begin(); iter != mPrevAxisValues.end(); iter++)
-		if (iter->second)
-			delete[] iter->second;
 
 	mPrevAxisValues.clear();
 
@@ -277,7 +278,11 @@ public:
 	std::string SDL_JoystickPathForIndex(int device_index)
 	{
 		if (m_JoystickPathForIndex != NULL)
-			return m_JoystickPathForIndex(device_index);
+		{
+			const char* path = m_JoystickPathForIndex(device_index);
+			if (path != NULL)
+				return std::string(path);
+		}
 
 		return "";
 	}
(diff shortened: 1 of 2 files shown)


Tartifless:
@lbrpdx : this is the commit i've been working on for the last 2 weeks, it fixes multiple reported crashes on windows and has been thoroughly tested with also multiple users that reported the issues.
It's mostly WIN32 except few null checks fixes.

I did not try linux compilation though.

PR #2204, opened 2026-08-29
WIN32: fix slow opening of menu

Tartifless:
Method canLocalUpdate() was added in batocera to enable update from USB key, it is useless for Windows.
==> This was added in may 2026

This commit overrides it in WIN32apisystems so that it does not trigger a cmd in Windows (RetroBat).

As a result, the update menu does not take multiple seconds to open anymore.


Tartifless:
@lbrpdx : can you merge this one ==> it's a bugfix for windows related with a method that was added in may 2026 for batocera.
It's a simple override in WIN32apisystem to ensure the method canLocalUpdate() directly returns false on windows.

PR #2206, opened 2026-09-02
fix: es config menu should display the audio from config, like for video

nadenislamarre:


nadenislamarre:
like the video and like what it saves.

PR #2207, opened 2026-09-03
feat: multi-user profiles UI (proof of concept — looking for design feedback)

Klutzleo:
## What this is

A working proof-of-concept for multi-user profile support in EmulationStation. We've been running this in production on OctaneOS (Radxa Cubie A7S / Allwinner A733) and it works well. We're opening this early to start a conversation about the right approach, not to request a merge.

## What it does

Adds a **Profiles** entry to the main ES menu that opens `GuiProfilesSettings`, which lets users:

- **Switch profiles** — each profile has its own RetroAchievements credentials (username, password, token, enabled/disabled) and its own save/state directories
- **Create new profiles** — named profiles stored under `/userdata/profiles/<name>/`
- **Rename profiles** — a picker lists all profiles; selecting one opens a keyboard to set a custom display name
- **Delete profiles** — with confirmation, deletes profile dir and switches back to default

**Display name priority:** custom `display_name` → RetroAchievements username → directory name. So profiles naturally show the RA handle without any manual setup.

**Save isolation:** `/userdata/saves` and `/userdata/states` are symlinks that repoint on every switch. configgen continues writing to `/userdata/saves` as always — the symlink transparently redirects to the active profile's directory. This was necessary because configgen hardcodes `SAVES = /userdata/saves` and overwrites `savefile_directory` at every game launch.

## What we know needs work

**ApiSystem integration:** The C++ currently calls `popen("batocera-profiles list ...")` and `system("batocera-profiles switch ...")` directly. We know these should move into `ApiSystem.cpp` as proper methods. We intentionally left this for v2 so we can get design feedback on the approach before doing the cleanup.

**Platform gating:** The backend script (`batocera-profiles`) uses `/userdata/profiles/` which is standard Batocera userdata layout, but we haven't added platform feature-flags yet.

**ES restart after switch:** RA widget on the main menu reflects the old session until ES restarts. We document this rather than solving it today.

## The backend

Profile management is handled by a `batocera-profiles` shell script (separate PR to batocera.linux if this direction is approved). It handles create/switch/delete, credential save/restore, and save directory symlink management.

## Running hardware

Tested on Radxa Cubie A7S with two real user profiles, wired and BLE controllers, SNES/GBA/NES saves, and RetroAchievements accounts. Works daily.

We'd love feedback on the overall approach before we invest in the ApiSystem refactor. Happy to iterate.

### DIFF
--- a/es-app/CMakeLists.txt
+++ b/es-app/CMakeLists.txt
@@ -87,7 +87,8 @@ set(ES_HEADERS
     ${CMAKE_CURRENT_SOURCE_DIR}/src/guis/GuiNetPlay.h # batocera
 	${CMAKE_CURRENT_SOURCE_DIR}/src/guis/GuiNetPlaySettings.h
     ${CMAKE_CURRENT_SOURCE_DIR}/src/guis/GuiRetroAchievements.h
-	${CMAKE_CURRENT_SOURCE_DIR}/src/guis/GuiRetroAchievementsSettings.h	
+	${CMAKE_CURRENT_SOURCE_DIR}/src/guis/GuiRetroAchievementsSettings.h
+	${CMAKE_CURRENT_SOURCE_DIR}/src/guis/GuiProfilesSettings.h
 	${CMAKE_CURRENT_SOURCE_DIR}/src/guis/GuiSystemInformation.h
 	${CMAKE_CURRENT_SOURCE_DIR}/src/guis/GuiBluetoothDevices.h
 	${CMAKE_CURRENT_SOURCE_DIR}/src/guis/GuiBluetoothPair.h
@@ -193,6 +194,7 @@ set(ES_SOURCES
 	${CMAKE_CURRENT_SOURCE_DIR}/src/guis/GuiNetPlaySettings.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/src/guis/GuiRetroAchievements.cpp
 	${CMAKE_CURRENT_SOURCE_DIR}/src/guis/GuiRetroAchievementsSettings.cpp
+	${CMAKE_CURRENT_SOURCE_DIR}/src/guis/GuiProfilesSettings.cpp
 	${CMAKE_CURRENT_SOURCE_DIR}/src/guis/GuiSystemInformation.cpp	
 	${CMAKE_CURRENT_SOURCE_DIR}/src/guis/GuiBluetoothDevices.cpp
 	${CMAKE_CURRENT_SOURCE_DIR}/src/guis/GuiBluetoothPair.cpp
--- a/es-app/src/guis/GuiMenu.cpp
+++ b/es-app/src/guis/GuiMenu.cpp
@@ -18,6 +18,7 @@
 #include "guis/GuiImageViewer.h"
 #include "guis/GuiNetPlaySettings.h"
 #include "guis/GuiRetroAchievementsSettings.h"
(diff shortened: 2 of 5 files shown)


Mikhailzrick:
Isn’t it precedent for consoles that different profiles have different save data? That’s what I would personally expect from having different profiles.

if switching profiles doesn’t at least cover separate saves data I would wonder what the point is since it’s such an integral part of having separate users/profiles.

This probably needs to tie into a select profile at boot system where paths can be determined then, and on profile switch, reload and rebuild whatever is needed.

symlinks anywhere in userdata is probably a non-starter also.

Klutzleo:
This does allow different save data and the users save states as well. 

The code allows:

- Switching of Retroachievements profiles and attaching the RA profile name to the profile.

- Create, rename, delete profiles. On creation you create your own name, never having need to touch RA if not desired. 

- All saves and Save states go into the profiles own folder. This has been tested. Saves and save states follow the user profile. You are unable to access the other users files in game. Save state slots and their hotkeys also carry over.

Klutzleo:
From what I saw, configgen hardcodes 

SAVES = /user data/saves

And overwrites 

savefile_directory

This is done on every single game launch. There is no config option to redirect it.

The symlink is the only mechanism that works transparently. Retroarch and configgen keep writing to

/user data/save

As always, but the symlink points that path at which ever profile is active. Both ends live in the same /userdata ext4 partition, so there is no cross partition issues.

If anyone can think of a different way I'm all ears, this is what I found works and in real time. No reboot necessary.

Mikhailzrick:
> From what I saw, configgen hardcodes
> 
> SAVES = /user data/saves
> 
> And overwrites
> 
> savefile_directory
> 
> This is done on every single game launch. There is no config option to redirect it.
> 
> The symlink is the only mechanism that works transparently. Retroarch and configgen keep writing to
> 
> /user data/save
> 
> As always, but the symlink points that path at which ever profile is active. Both ends live in the same /userdata ext4 partition, so there is no cross partition issues.
> 
> If anyone can think of a different way I'm all ears, this is what I found works and in real time. No reboot necessary.

Maybe an environment variable or something. 

Userdata/profile/saves for example.

though this kind of change is going to make upgrading / downgrading potentially broken/messy

Mikhailzrick:
Userdata could be NTFS or a NAS, etc. something to consider. There’s no guarantee there symlinks will be available.

Klutzleo:
Absolutely. Let me do some more brainstorming on different storage options and upgrading and downgrading. Thanks for the thoughts, this is great!

Klutzleo:
What if we patch batoceraPaths.py in configgen to read the active profile directly instead of using symlinks? Something like:


_profile = open('/userdata/system/.batocera-profiles.current').read().strip()
_dir = _profile if _profile else '.default'
SAVES = Path('/userdata/profiles') / _dir / 'saves'
configgen runs fresh on every game launch so it'd pick up the right profile every time. Would also work on NTFS/NAS since there are no symlinks involved.

Mikhailzrick:
Bryan would be the one to ask about that considering the recent huge change to the launcher and configgens. But it should be do-able.

also needs to be discussed exactly where in the directory tree the profile variable should be.
Userdata/profile/ would be the easiest/cleanest maybe. But that also means separated gameslist/roms list among everything else.

if we want it more narrowly scoped, maybe just userdata/saves/profile and userdata/system/comfigs/profile etc

PR #2210, opened 2026-09-05
[WIN32] Fix crash on virtual mouse devices

Tartifless:
Impacts only if WIN32 sections.

1) Ensure vmulti is not always classified as lightgun (only when wiimotegun is running) 2) Fix buffer issue in GetRawInputDevicePath
3) Fix truncated names in GetRawInputDeviceName
4) Fix potential crash due to insufficient buffer if rawdevice list changes due to potential HID virtual device disconnection 5) Gun creation loop should consume devicePaths[i] instead of calling the system again 6) Protect nullptr risk
7) Added a diagnostic trace

### DIFF
--- a/es-core/src/GunManager.cpp
+++ b/es-core/src/GunManager.cpp
@@ -10,6 +10,7 @@
 #include "LocaleES.h"
 #include "renderers/Renderer.h"
 #include "InputManager.h"
+#include <vector>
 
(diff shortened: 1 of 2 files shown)


Tartifless:
@lbrpdx : can you merge this ? (WIN32 only)

PR #2211, opened 2026-09-07
[WIN32] Fix windows build + differentiate both controllers in case of multiplayer board with same path

Tartifless:


Tartifless:
@nadenislamarre @lbrpdx : please merge this one, commit https://github.com/batocera-linux/batocera-emulationstation/commit/e401b9a950a4ce9b8fe6134cc7d57f94b4e230bd has broken windows build

PR #2212, opened 2026-09-09
[WIN32] Prioritize a gamecontrollerdb file over internal mapping

Tartifless:


Tartifless:
@lbrpdx : WIN32 only, small change to prioritize a gamecontrollerdb file to get the mapping over the internal sdl2 database if the file exists.

